PR #11, opened 2025-09-28
Add vectored I/O (readv/writev) support to libxev

lalinsky:
## Summary

This PR adds comprehensive vectored I/O support to libxev, enabling efficient scatter-gather operations across all supported backends.

## Changes

- **Add `vectors` variant to ReadBuffer and WriteBuffer unions**
- **Implement vectored I/O in all 5 backends:**
  - **io_uring**: Uses `prep_readv`/`prep_writev` for optimal performance
  - **epoll**: Uses `posix.readv`/`writev`/`preadv`/`pwritev`
  - **kqueue**: Uses `posix.readv`/`writev`/`preadv`/`pwritev`
  - **IOCP**: Uses `WSASend`/`WSARecv` with multiple `WSABUF` structures
  - **wasi_poll**: Uses native iovec support
- **Update all high-level watchers** (File, Stream, TCP, UDP) to support vectors
- **Add proper partial write handling** for vectored operations
- **Add comprehensive test coverage** for vectored file I/O
- **Dynamic API intentionally excludes vectors** (common subset design)

## Features

✅ **Unified API across platforms**  
✅ **Zero-copy operations**  
✅ **Optimal backend-specific implementations**  
✅ **Backward compatible**  
✅ **Comprehensive test coverage**  
✅ **Proper partial write handling**  

## Usage Example

```zig
// Create vectors pointing to your data
var vectors = [_]std.posix.iovec_const{
    .{ .base = buf1.ptr, .len = buf1.len },
    .{ .base = buf2.ptr, .len = buf2.len },
    .{ .base = buf3.ptr, .len = buf3.len },
};

// Use with any libxev watcher
file.write(&loop, &completion, .{ .vectors = &vectors }, ...);
tcp.write(&loop, &completion, .{ .vectors = &vectors }, ...);
udp.send(&loop, &completion, .{ .vectors = &vectors }, ...);
```

## Test Plan

- [x] All backends compile successfully
- [x] Vectored file I/O test passes
- [x] Existing tests remain unaffected
- [x] Partial write handling works correctly
- [x] Dynamic API properly excludes vectors

## Performance Benefits

Vectored I/O provides significant performance benefits for applications that need to:
- Send/receive data from multiple non-contiguous buffers
- Avoid memory copying operations
- Implement efficient protocol serialization
- Handle scatter-gather network operations

The implementation uses the most efficient vectored I/O mechanism available on each platform while providing a unified high-level API.

### DIFF
--- a/vendor/libxev/src/backend/epoll.zig
+++ b/vendor/libxev/src/backend/epoll.zig
@@ -979,6 +979,7 @@ pub const Completion = struct {
                 const n_ = switch (op.buffer) {
                     .slice => |v| posix.read(op.fd, v),
                     .array => |*v| posix.read(op.fd, v),
+                    .vectors => |v| posix.readv(op.fd, v),
                 };
 
                 break :res .{
@@ -993,6 +994,7 @@ pub const Completion = struct {
                 const n_ = switch (op.buffer) {
                     .slice => |v| posix.pread(op.fd, v, op.offset),
                     .array => |*v| posix.pread(op.fd, v, op.offset),
+                    .vectors => |v| posix.preadv(op.fd, v, op.offset),
                 };
 
                 break :res .{
@@ -1007,6 +1009,7 @@ pub const Completion = struct {
                 .write = switch (op.buffer) {
                     .slice => |v| posix.write(op.fd, v),
                     .array => |*v| posix.write(op.fd, v.array[0..v.len]),
+                    .vectors => |v| posix.writev(op.fd, v),
                 },
             },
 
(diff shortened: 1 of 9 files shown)


lalinsky:
@claude consider what would we need in order to change the API to slices instead of iovecs. So vectors would become slice of slice of u8. What would be the implications. Review the approach. Suggest implemention path.

lalinsky:
@claude explain the windows case to me, currently we use iovecs, which are zero copy on non-windows, copy on windows, the change would make it copy on all platforms, but I don't see any dual conversion for windows

PR #33, opened 2025-10-07
Implement task recycling with 512-byte dynamic layout

### LINKED ISSUE #31: Recycle tasks
- reserve 256 bytes for result data in AnyTask
- in Task(T), construct FutureResult(T) on demand by casting the data
- use MemoryPool to recycle AnyTask objects

lalinsky:
Closes #31

## Summary

Implements task recycling for improved performance using a 512-byte dynamic layout approach:

- Create TaskPool that allocates fixed 512-byte blocks with 16-byte alignment
- Both AnyTask and AnyBlockingTask can be allocated from the pool
- FutureResult(T) is dynamically placed after the base struct using pointer arithmetic
- For BlockingTask, Args are also dynamically placed after FutureResult(T)
- Large blocking tasks (>512 bytes) fall back to regular allocator
- Time-based retention (60s default) with configurable warm pool minimum (4 default)
- TaskPool cleanup runs alongside StackPool cleanup in runtime loop

## Benefits

- Reduced allocation pressure for coroutine-based tasks
- BlockingTask can now be pooled (when Args fit in 512 bytes)
- Cache-friendly contiguous layout for task data
- Automatic fallback for oversized blocking tasks
- Memory efficiency with time-based cleanup

All tests pass!

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/runtime.zig
+++ b/src/runtime.zig
@@ -15,6 +15,9 @@ const FutureResult = @import("future_result.zig").FutureResult;
 const stack_pool = @import("stack_pool.zig");
 const StackPool = stack_pool.StackPool;
 const StackPoolOptions = stack_pool.StackPoolOptions;
+const task_pool = @import("task_pool.zig");
+const TaskPool = task_pool.TaskPool;
+const TaskPoolOptions = task_pool.TaskPoolOptions;
 
 // Compile-time detection of whether the backend needs ThreadPool
 fn backendNeedsThreadPool() bool {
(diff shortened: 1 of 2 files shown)


lalinsky:
@claude maybe we need 1024b for the tasks

lalinsky:
@claude I think we could use anonymous structs like this for better field access that automatically deals with field alignment:

```
            const ExtraData = struct {
                task: AnyTask,
                result: FutureResult(T),
                args: Args,  
            };

     const extra_data: *ExtraData = @ptrCast(&self.any_task);
     return extra_data.result;
```

This is assuming the order of fields in struct is guaranteed.

PR #38, opened 2025-10-09
Add BroadcastChannel synchronization primitive

lalinsky:
## Summary

Implements a non-blocking broadcast channel similar to Tokio's `broadcast::channel`, allowing multiple consumers to each receive all broadcast messages with automatic handling of slow consumers.

## Features

- **Non-blocking producers**: `send()` never blocks, overwriting oldest values when the buffer is full
- **Dynamic subscription**: Consumers can subscribe/unsubscribe at any time using an intrusive linked list (zero allocations)
- **Lag detection**: Slow consumers automatically receive `error.Lagged` when they fall more than `buffer.len` behind, then skip to the oldest available message
- **Closeable**: Channel can be closed to prevent new sends while allowing consumers to drain remaining messages
- **Ring buffer**: Efficient fixed-capacity implementation with monotonic write position tracking

## API

```zig
var buffer: [10]Message = undefined;
var channel = BroadcastChannel(Message).init(&buffer);

// Producer
try channel.send(runtime, message);
channel.close(runtime);

// Consumer
var consumer = BroadcastChannel(Message).Consumer{};
channel.subscribe(runtime, &consumer);
defer channel.unsubscribe(runtime, &consumer);

while (true) {
    const msg = channel.receive(runtime, &consumer) catch |err| switch (err) {
        error.Lagged => continue, // Skip and get next available
        error.Closed => break,
        else => return err,
    };
    // Process message
}
```

## Testing

Comprehensive test coverage includes:
- Basic send/receive operations
- Multiple consumers receiving same messages
- Lag detection and recovery
- Non-blocking `tryReceive()`
- New subscriber positioning
- Unsubscribe isolation
- Close semantics (drain, wake waiters, prevent new sends)

All tests pass.

## Summary by CodeRabbit

- New Features
  - Adds a broadcast channel for fan-out messaging with a fixed-capacity ring buffer.
  - Supports multiple subscribers with per-consumer read positions and a Consumer handle for subscription management.
  - Exposes subscribe/unsubscribe, blocking receive, non-blocking tryReceive, send, and close APIs; handles overwrite/lag semantics and is single-runtime only.

- Tests
  - Adds comprehensive tests for multi-consumer delivery, lag behavior, subscription/unsubscription, new-subscriber semantics, close/drain behavior, wake-on-close, and non-blocking receive on closed channels.

### DIFF
--- a/src/sync.zig
+++ b/src/sync.zig
@@ -1408,18 +1408,20 @@ test "BroadcastChannel: basic send and receive" {
 
     var buffer: [10]u32 = undefined;
     var channel = BroadcastChannel(u32).init(&buffer);
+    var receiver_ready = ResetEvent.init;
 
     const TestFn = struct {
-        fn sender(rt: *Runtime, ch: *BroadcastChannel(u32)) !void {
-            rt.yield(); // Let receiver subscribe first
+        fn sender(rt: *Runtime, ch: *BroadcastChannel(u32), ready_event: *ResetEvent) !void {
+            ready_event.wait(rt); // Wait for receiver to subscribe
             try ch.send(rt, 1);
             try ch.send(rt, 2);
             try ch.send(rt, 3);
         }
 
-        fn receiver(rt: *Runtime, ch: *BroadcastChannel(u32), consumer: *BroadcastChannel(u32).Consumer, results: *[3]u32) !void {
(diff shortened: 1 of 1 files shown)


lalinsky:
@claude try to investigate why are the new tests hanging

lalinsky:
@claude use ResetEvent for synchronization, not depending and hoping that yield will help

PR #43, opened 2025-10-10
Add signal handling support

### LINKED ISSUE #35: Signal handlers
We should add support for signal handlers. After installing a handler, we should start a coroutine watching a queue for signal events. If an event comes, we call the user callback (in the coroutine context). That way, the user can do anything inside the signal handler, even wait on locks or other operations.

lalinsky:
Implements cross-platform signal handling as described in #35.

## Implementation
- Single coroutine handles all signals via atomic counter array
- Signal handlers increment counters and trigger xev.Async
- User callbacks run in coroutine context (can use locks, I/O, etc.)
- Cross-platform: sigaction on Unix, SetConsoleCtrlHandler on Windows

Closes #35

----
Generated with [Claude Code](https://claude.ai/code)

## Summary by CodeRabbit

* **New Features**
  * Added cross-platform signal handling (INT, TERM, HUP, QUIT, USR1, USR2) integrated with the runtime, letting apps register callbacks invoked on signal events.
  * Exposed signal types and a SignalHandler via the public API for easy integration and lifecycle management.
* **Documentation**
  * Added an example demonstrating graceful shutdown, status output, and clean resource teardown.

### DIFF
--- a/examples/signal_demo.zig
+++ b/examples/signal_demo.zig
@@ -4,13 +4,13 @@ const zio = @import("zio");
 var should_exit = std.atomic.Value(bool).init(false);
 var signal_count = std.atomic.Value(u32).init(0);
 
-fn handleShutdown() void {
-    const count = signal_count.fetchAdd(1, .monotonic) + 1;
+fn handleShutdown(args: struct { should_exit_ptr: *std.atomic.Value(bool), signal_count_ptr: *std.atomic.Value(u32) }) void {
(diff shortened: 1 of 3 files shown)


lalinsky:
@claude

1) fix windows build

2) I'd like this to be internal to Runtime, on runtime we just have addSignalHandler function that spawns the internal coroutine as needed

3) I'd like the signal handlers to receive args, we can reserve a small space, e.g. 64 bytes on the handler struct and comptime assert that the args tuple can fit that

PR #58, opened 2025-10-19
Add CompactConcurrentQueue (8-byte variant) with benchmarks

### LINKED ISSUE #57: 8 byte large concurrent queue
I'd like to create a variant of ConcurrentQueue, that only stores the head, and the tail is stored in the first item, that would need a field in the item to support this, but I think it could work. We need this because we only get 64bit state from std.Io.Mutex and Condition 

@claude evaluate this and consider if it's it's possible

lalinsky:
## Summary

Implements a space-optimized version of ConcurrentQueue that fits in 8 bytes for compatibility with `std.Io.Mutex` and `Condition` 64-bit state.

**Key features:**
- Queue structure: 8 bytes (only atomic head pointer)
- Tail pointer stored in head node's `tail` field
- Same O(1) operations as standard ConcurrentQueue
- Comprehensive tests and benchmark suite

**Trade-offs:**
- ✅ Fits in 64-bit atomic word
- ✅ Maintains all functionality
- ❌ Each node 8 bytes larger (tail field)
- ❌ Slightly more complex operations

Closes #57

---

Generated with [Claude Code](https://claude.ai/code)

## Summary by CodeRabbit

* **New Features**
  * Introduced CompactConcurrentQueue — an 8-byte, memory-efficient concurrent queue.
  * Added comprehensive benchmark suite with single- and multi-threaded performance tests.

* **Updates**
  * Synchronization primitives (mutexes/conditions) now use the compact queue to reduce per-waiter memory overhead.

* **Chores**
  * Exposed ConcurrentQueue and CompactConcurrentQueue in the public API.

### DIFF
--- a/src/core/WaitNode.zig
+++ b/src/core/WaitNode.zig
@@ -7,6 +7,7 @@ vtable: *const VTable,
 // For participation in wait queues
 prev: ?*WaitNode = null,
 next: ?*WaitNode = null,
+tail: ?*WaitNode = null, // For CompactConcurrentQueue
 in_list: if (builtin.mode == .Debug) bool else void = if (builtin.mode == .Debug) false else {},
 
 pub const VTable = struct {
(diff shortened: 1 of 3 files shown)


lalinsky:
@claude use this variant in mutex and condition

PR #69, opened 2025-10-23
Add Notify sync primitive

### LINKED ISSUE #68: Add Notify sync primitive
I'd like a sync primitive like ResetEvent, but signal-triggered. It would have methods wait, signal and broadcast. It should be implement using WaitQueue, should be fairly simple.

@claude plan this

lalinsky:
## Summary

Implements a signal-triggered synchronization primitive that allows tasks to wait for transient signals.

## Changes

- **New file**: `src/sync/Notify.zig` - Complete implementation with comprehensive docs and tests
- **Updated**: `src/sync.zig` - Export Notify and add test reference
- **Updated**: `README.md` - Add Notify to sync primitives list

## Features

- `signal()`: Wakes one waiting task (FIFO order)
- `broadcast()`: Wakes all waiting tasks
- `wait()`: Suspends until signaled
- `timedWait()`: Suspends with timeout
- Signals are lost if no tasks are waiting (no-op)

## Implementation

Uses WaitQueue with only sentinel0 state (no persistent "signaled" state). Follows the ResetEvent pattern but simpler due to stateless design.

All tests pass successfully.

Closes #68

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ the interface, or deprecate this library.
 - File I/O on all platforms, Linux and Windows are truly non-blocking, other platforms are simulated using a thread pool
 - Network I/O, supports TCP/UDP sockets, DNS resolution currently via thread pool
 - Full `std.Io.Reader` and `std.Io.Writer` support for TCP streams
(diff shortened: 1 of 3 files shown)


lalinsky:
@coderabbitai review

PR #83, opened 2025-10-26
Fix handling of error.EOF in net.Stream.Reader

lacc97:
This fixes an issue I was having in which net.Stream.Reader.streamRemaining() was failing because it wasn't recognising the end of stream condition correctly. The behaviour now matches the Stream.Reader in src/stream.zig as well.

Should I add a test for this in src/io/test/net.zig?

### DIFF
--- a/src/io/test/net.zig
+++ b/src/io/test/net.zig
@@ -156,6 +156,49 @@ pub fn checkBind(server_addr: anytype, client_addr: anytype) !void {
     try runtime.runUntilComplete(Test.mainFn, .{ runtime, server_addr, client_addr }, .{});
 }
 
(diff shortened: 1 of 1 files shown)


lalinsky:
Oh, thank you. I was refactoring these network calls recently and completely missed this translation.

Yeah, it would be great to add a test that calls shutdown on the socket on one end, and expects the correct error on the other end.

But it's OK if you don't add it, I can do it.

lacc97:
Sure I can add a test.

lacc97:
Added a test to check that EOF return error.EndOfStream at the std.Io.Reader interface. I also checked the code before my changes failed the test (at least it did on my machine :).

lalinsky:
Perfect. Thank you!

PR #89, opened 2025-10-27
Refactor task state

lalinsky:


lalinsky:
@coderabbitai explain the lost-race yield issue to me?

lalinsky:
@coderabbitai is there a place in our code where we don't use preparing_to_wait and should use it?

lalinsky:
@claude apply the fix from the last comment

PR #94, opened 2025-10-28
Fix cancellation race in Channel, Condition, and Notify

lalinsky:
Fixes a race condition where waiters could be removed from queues by wakers (send/signal) but then cancelled before wake() is called, resulting in lost wakeups and stuck resources.

The bug occurred in this sequence:
1. Waker pops wait_node from queue
2. Waiter gets cancelled
3. Waiter tries to remove from queue -> returns false (already removed)
4. Waiter ignores the return value and returns error.Canceled
5. Waker calls wake() but waiter is already gone
6. Resource/signal is lost - no other waiter is notified

Impact:
- Channel: Lost items in buffer or lost buffer slots
- Condition: Lost signals meant for one waiter
- Notify: Lost notifications

The fix checks the return value of remove() in cancellation handlers. If remove() returns false (waiter was already removed by waker), we wake another waiter to consume the resource/signal instead.

Primitives fixed:
- Channel.receive() and Channel.send()
- Condition.wait()
- Notify.wait() and Notify.timedWait()

Primitives already correct:
- Mutex.lock() (already unlocks to pass lock)
- Semaphore.wait() (re-signals at higher level)
- ResetEvent.wait() (level-triggered, state persists)

### DIFF
--- a/src/sync/Notify.zig
+++ b/src/sync/Notify.zig
@@ -228,7 +228,15 @@ pub fn asyncWait(self: *Notify, wait_node: *WaitNode) bool {
 /// Cancels a pending wait operation by removing the wait node.
 /// This is part of the Future protocol for select().
 pub fn asyncCancelWait(self: *Notify, wait_node: *WaitNode) void {
-    _ = self.wait_queue.remove(wait_node);
+    const was_in_queue = self.wait_queue.remove(wait_node);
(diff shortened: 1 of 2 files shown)


lalinsky:
@coderabbitai review

lalinsky:
@coderabitai review

PR #96, opened 2025-10-28
Fix net.Stream unbuffered writes

lacc97:
This PR adds a check to skip adding empty buffers to the IO vector for net.Stream.

The missing check for an empty buffer in `addBuf` meant that for an unbuffered Writer, a zero-length buffer was unconditionally added to the vector sent to libxev. On the linux io_uring backend, this caused the send syscall to fail with `EFAULT` due to the pointer value being the undefined pattern (0xaa..).

I also added a test for this by copying and modifying the `checkListen()` helper in io/test/net.zig but I wonder if it would be better to just add a flag on the original `checkListen()` function instead.

### DIFF
--- a/src/io/net.zig
+++ b/src/io/net.zig
@@ -698,17 +698,16 @@ pub fn netRead(rt: *Runtime, fd: Handle, bufs: [][]u8) !usize {
 
 fn addBuf(buf: *xev.WriteBuffer, data: []const u8) !void {
     if (data.len == 0) return;
-    if (buf.vectors.len < buf.vectors.data.len) {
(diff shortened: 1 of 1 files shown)


lacc97:
I ended up modifying the checkListen() helper for the test instead which duplicates less code.

I also fixed the thing the rabbit found in https://github.com/lalinsky/zio/pull/96#pullrequestreview-3390956790 by returning early if the buffer is full. It was an inefficiency in which the io vector wasn't used to it's capacity, not a correctness issue per se.

lalinsky:
Thank you, looks good.

PR #97, opened 2025-10-29
Optimize select() winner identification with atomic tracking

lalinsky:
Replace runtime iteration over signaled flags with direct winner index lookup.

Changes:
- SelectWaiter now tracks shared atomic winner index and own index
- First future to complete claims winner slot via cmpxchg
- Remove per-waiter signaled flags and wake counter
- Add NO_WINNER sentinel constant for unclaimed state

Benefits:
- Direct winner index lookup (no runtime conditionals per waiter)
- Race-free atomic winner claiming
- Only first winner wakes parent (no spurious wakeups)
- Simpler code with fewer atomic operations per waiter

All tests pass (150/150).

### DIFF
--- a/src/sync/channel.zig
+++ b/src/sync/channel.zig
@@ -382,6 +382,11 @@ pub fn AsyncReceive(comptime T: type) type {
             // Perform the receive operation under lock
             self.channel.mutex.lock();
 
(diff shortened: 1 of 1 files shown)


lalinsky:
@coderabbitai what do you think about this change?

PR #98, opened 2025-10-30
Add Signal support to select()

lalinsky:
## Summary

Implements the Future protocol for `Signal` to enable it to be used with `select()`. This allows waiting on multiple signals or mixing signals with other async operations like tasks and channels.

### Core Implementation
- **Future protocol declarations**: Added `Result = void` and `WaitContext` struct
- **asyncWait()**: Implements fast path for already-received signals, registers xev async wait
- **waitCallback()**: Simple callback that wakes parent wait node
- **asyncCancelWait()**: Cancels xev operation if still active, prevents spurious wakes
- **getResult()**: Consumes signal counter

### Refactoring
- Unified `wait()` and `timedWait()` to use the same `WaitContext` and `waitCallback()`
- Removed duplicate local context structs and inline callbacks
- Simplified to use `parent_wait_node.wake()` instead of direct task resumption
- Removed unused imports (`AnyTask`, `resumeTask`)

### Bug Fixes
- Fixed `select.zig` `default_value_ptr` field for Zig 0.15 compatibility

## Tests

Added 3 comprehensive tests:
- `Signal: select on multiple signals` - Wait on SIGUSR1 vs SIGUSR2
- `Signal: select with signal already received (fast path)` - Pre-received signal returns immediately
- `Signal: select with signal and task` - Mix signal with slow task

All 153 tests pass.

## Example Usage

```zig
var sig1 = try Signal.init(.interrupt);  // SIGINT
defer sig1.deinit();
var sig2 = try Signal.init(.terminate);  // SIGTERM
defer sig2.deinit();

const result = try select(rt, .{
    .interrupt = &sig1,
    .terminate = &sig2,
});

switch (result) {
    .interrupt => std.debug.print("Received SIGINT\n", .{}),
    .terminate => std.debug.print("Received SIGTERM\n", .{}),
}
```

### DIFF
--- a/src/signal.zig
+++ b/src/signal.zig
@@ -396,9 +396,7 @@ pub const Signal = struct {
             const cancelIo = @import("io/base.zig").cancelIo;
             cancelIo(rt, &ctx.completion);
         } else {
-            // The xev operation already completed, which means the signal was received
-            // but we're cancelling because another future won the select race.
-            // Notify the event again so another waiter can be woken.
+            // Signal was delivered but not consumed; wake another waiter to handle it.
             self.entry.event.notify() catch {};
         }
     }


lalinsky:
@claude I think there is a cancelation safety issue, if we have signal triggering, but not winning the select race, we might need to call event.notify() to potentially wake up another one

lalinsky:
@claude address the last review comment

Consider adding an inline comment at line 399 explaining why notify() is needed: "Signal was delivered but not consumed; wake another waiter to handle it."

PR #134, opened 2025-11-06
[event-loop] Kqueue backend

lalinsky:
WIP

### DIFF
--- a/src/event/backends/kqueue.zig
+++ b/src/event/backends/kqueue.zig
@@ -42,6 +42,7 @@ const EV_EOF: u16 = 0x8000;
 allocator: std.mem.Allocator,
 kqueue_fd: i32 = -1,
 async_impl: ?AsyncImpl = null,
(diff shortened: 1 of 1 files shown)


lalinsky:
@claude help me finish this

lalinsky:
@claude you should use the kqueue definitions from std.c, not define your own

lalinsky:
@claude do we need the poll_queue? does kqueue also have problem with having duplicate fds in there? or can we create e.g. one event for read, one for write. if that's possible, we should not use the queue.

lalinsky:
@claude do the refactoring to remove the poll queue

lalinsky:
@claude we should use EVFILT_USER for async

lalinsky:
@claude What do we need to do to fix freebsd?

lalinsky:
@claude this is incredibly inefficient, one syscall per operations, submissions need to be batches, we submit one batch, get one batch, proceed until we submitted all requests we had when we started tick, only the last all is blocking

PR #141, opened 2025-11-13
fix small typos

dacheng-gao:


lalinsky:
Hehe, I was confusing myself here. i wanted to use cancelled all the way, but the noticed std used the american canceled. It seems like cancellation is the preferred variant in both.

dacheng-gao:
Thanks for merging.

Yeah. It confused me a lot too.
And I catch these typos with help of IDE (CLion).

PR #159, opened 2025-12-14
Allow async operations from main thread without spawning

lalinsky:
## Summary

- Add `main_task` to Executor to support calling `rt.sleep()` and other async operations directly from the main thread without requiring `spawn()`
- Unify `run()` and `runSchedulerLoop()` - both now use the same scheduler loop with different exit conditions
- Remove dead code: `ExecutorWaiter`, `ThreadWaiter`, and related no-task handling in select

## Test plan

- [x] All 168 existing tests pass
- [x] New test "Runtime: sleep from main" validates the feature


lalinsky:
@coderabbitai review

PR #169, opened 2025-12-22
examples: Consistently use rt instead of runtime

perillo:
In the examples, the use of the runtime variable in the main function inconsistently use either `rt` or `runtime`.

In the docs, instead, the use is always `rt`.

This change will use `rt` consistently, but only for examples and documentation.


lalinsky:
Thank you!

PR #206, opened 2026-01-16
Remove ready_queue_next

lalinsky:
Increase of trying to invent the concept of a tick, we will just run tasks that are ready, until we either reach the end, or we already ran 61 of them.

This makes is easier to implement work-stealing later on, it also improves CPU-bound tasks. It might slightly increase latency for I/O tasks, but if there are I/O tasks, they will simple clear the ready queue and do poll more often.

The worse case are mixed workloads, I/O latency could increase, but we will add more control over this and work-stealing will also help with balancing the load.


lalinsky:
@coderabbitai can you guess the reason behind this? if so, does it make sense to yuo?

PR #207, opened 2026-01-16
Rename Timeout to AutoCancel

lalinsky:
## Summary
- Rename `zio.Timeout` to `zio.AutoCancel` to free up the `Timeout` name for a future tagged union type (matching `std.Io.Timeout`)
- Move timeout consumption logic to `Awaitable.consumeAutoCancel()`
- Replace `Runtime.checkTimeout()` with `AutoCancel.check()` returning bool
- Rename `CanceledStatus.timeout` to `auto_canceled`

## Test plan
- [x] All existing tests pass
- [x] Full build (examples and benchmarks) succeeds

### DIFF
--- a/src/runtime/autocancel.zig
+++ b/src/runtime/autocancel.zig
@@ -35,8 +35,11 @@ pub const AutoCancel = struct {
     }
 
     pub fn set(self: *AutoCancel, rt: *Runtime, timeout: Duration) void {
-        // Skip setting timer if waiting forever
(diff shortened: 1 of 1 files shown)


lalinsky:
@coderabbitai pending_errors should NOT actually be decreased here, the error was already delivered, that's why it has err as parameter, to make it obvious to the user to only call this after they received error.Canceled

lalinsky:
@coderabbitai you seem to be misunderstanding how the CancelationStatus works

PR #223, opened 2026-01-21
io_uring: handle EINTR by resubmitting operations

lalinsky:
## Summary

- io_uring can return -EINTR in CQE results when operations are interrupted by signals
- Instead of propagating this as an unexpected error, resubmit the operation automatically
- Make `submit()` handle both initial submission and resubmission to avoid code duplication

## Test plan

- [x] All existing tests pass


lalinsky:
@coderabbitai review

PR #273, opened 2026-02-03
Refactor current_coroutine tracking to eliminate special cases

lalinsky:
## Summary

Refactors `current_coroutine` tracking to eliminate special-case handling in `getCurrentTask()`.

Previously, `current_coroutine` was `null` when in the main task context, requiring a special case check. Now it always points to the active coroutine:

- **Main task context**: `current_coroutine = &main_task.coro`
- **Inside `run()` scheduler**: `current_coroutine = null`
- **Running spawned task**: `current_coroutine = &task.coro`

## Changes

- **`Executor.init()`**: Initialize `current_coroutine` to `&main_task.coro`
- **`Executor.run()`**: Set to `null` on entry, restore to `&main_task.coro` on exit (via `defer`)
- **Task stepping**: Set to `null` immediately after `step()` returns, before callbacks
- **`Executor.yield()`**: Check `is_main` by comparing to `&main_task.coro` instead of `null`
- **`Executor.getCurrentTask()`**: Simplified to always use `current_coroutine` (no special case)

## Benefits

1. **Cleaner API**: No special cases in `getCurrentTask()`
2. **Better semantics**: Cleanup callbacks always run in scheduler context (`current_coroutine = null`)
3. **More explicit**: The transition between task and scheduler contexts is now clearly visible

## Test plan

- [x] All existing tests pass (`./check.sh`)

### DIFF
--- a/src/runtime.zig
+++ b/src/runtime.zig
@@ -253,7 +253,7 @@ pub const Executor = struct {
 
     id: u6,
     loop: ev.Loop,
-    current_coroutine: ?*Coroutine = null,
+    current_task: ?*AnyTask = null,
 
     ready_queue: SimpleQueue(WaitNode) = .{},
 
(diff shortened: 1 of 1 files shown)


lalinsky:
@claude i think we could rename current_coroutine to current_task

PR #307, opened 2026-02-09
Add blocking socket operations support

### LINKED ISSUE #288: Implement blocking socket ops

lalinsky:
### Summary

Implements blocking socket operations following the pattern from PR #306 (blocking pipe operations).

### Implementation

- **POSIX**: Poll for readiness with infinite timeout, then perform I/O
- **Windows**: Direct blocking I/O

### Operations Added

- recv/send for stream sockets
- recvfrom/sendto for datagram sockets
- recvmsg/sendmsg for advanced message operations
- close/shutdown for connection management

All 355 unit tests pass successfully.

Fixes #288

### DIFF
--- a/src/ev/blocking.zig
+++ b/src/ev/blocking.zig
@@ -22,6 +22,9 @@ const NetRecvMsg = @import("completion.zig").NetRecvMsg;
 const NetSendMsg = @import("completion.zig").NetSendMsg;
 const NetOpen = @import("completion.zig").NetOpen;
 const NetBind = @import("completion.zig").NetBind;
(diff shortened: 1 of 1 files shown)


lalinsky:
@claude the windows check needs to be outside the poll func, it fails on windows

lalinsky:
@claude but we could actually use poll for net recv send ops on windows, that would allow us to support timeouts later on

lalinsky:
@claude what would be needed to support listen/accept/connect as well?

lalinsky:
@claude implement it

PR #341, opened 2026-03-12
Add native DNS resolver

lalinsky:
## Summary

- Pure Zig DNS resolver that doesn't depend on libc for name resolution
- Supports UDP/TCP queries with automatic TCP fallback for truncated responses
- Parses /etc/resolv.conf with automatic reload (checks mtime every 5 seconds)
- Supports /etc/hosts lookup before DNS queries
- Handles search domains, multiple nameservers, and server rotation
- Build option `-Ddns-resolver=native` to select resolver backend

## Test plan

- [x] `./check.sh` passes
- [ ] Manual testing with `-Ddns-resolver=native`

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -13,9 +13,16 @@ pub fn build(b: *std.Build) void {
         "Override the default event loop backend (io_uring, epoll, kqueue, iocp, poll)",
     );
 
+    const dns_resolver = b.option(
+        []const u8,
+        "dns-resolver",
+        "Override the default DNS resolver (system, native)",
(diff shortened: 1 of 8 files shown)


lalinsky:
@claude read reviews, evaluate and fix what needs fixing

lalinsky:
@coderabbitai full review

lalinsky:
@claude we should not be using std.fs in this code

lalinsky:
@claude hosts file is not cached?

PR #344, opened 2026-03-13
Remove preparing_to_wait state, use awaken bit instead

### LINKED ISSUE #343: Remove preparing_to_wait state
Let's explore the possibility of removing the preparing_to_wait state and instead using unparked/awaken flag in a packed bit status, similarly to NetBSD's park/unpark primitives.

lalinsky:
Closes #343

Replace the `preparing_to_wait` task state with a NetBSD-style park token (`awaken` bit) packed into the task state byte.

Previously, `preparing_to_wait` bridged the race window between when `waitTask` arms and the actual context switch. Wakers that saw `preparing_to_wait` would transition to `.ready` without queuing the task.

The new approach packs state and a boolean `awaken` flag into a single `u8`. When a waker fires while the task is in `.ready` state, it sets `awaken=true` via atomic CAS instead of returning as a no-op. `processCleanup.park` then consumes the token and reschedules the task instead of transitioning to `.waiting`.

This also simplifies `waitTask`: the `preparing_to_wait` store and double-check re-arm loop are no longer needed.

### DIFF
--- a/src/runtime.zig
+++ b/src/runtime.zig
@@ -1201,3 +1201,66 @@ test "Runtime: multi-threaded with task migration" {
 
     try std.testing.expectEqual(100, ctx.counter.load(.acquire));
 }
+
+test "runtime: wake-before-park awaken bit stress (single executor)" {
+    try wakeBeforeParkStress(1);
+}
+
+test "runtime: wake-before-park awaken bit stress (two executors)" {
+    try wakeBeforeParkStress(2);
+}
+
+fn wakeBeforeParkStress(executor_count: u6) !void {
+    const ResetEvent = @import("sync/ResetEvent.zig");
+
+    const runtime = try Runtime.init(std.testing.allocator, .{
+        .executors = .exact(executor_count),
+    });
+    defer runtime.deinit();
+
+    const Ctx = struct {
+        // Number of ping-pong iterations to exercise the wake-before-park window.
+        const iterations: u32 = 10_000;
+
+        ping: ResetEvent = .{},
+        pong: ResetEvent = .{},
+        counter: std.atomic.Value(u32) = .init(0),
+
+        // Waits on ping each iteration — this is the task that parks, and the one
+        // whose awaken bit gets set when the waker fires between the condition check
+        // and the actual park CAS in processCleanup.park.
+        fn parker(ctx: *@This()) !void {
+            for (0..iterations) |_| {
+                try ctx.ping.wait();
+                ctx.ping.reset();
+                _ = ctx.counter.fetchAdd(1, .release);
+                ctx.pong.set();
+            }
+        }
+
+        // Fires ping immediately each iteration, without waiting for parker to park first.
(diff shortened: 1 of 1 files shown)


src/task.zig
```diff
@@ -157,7 +157,7 @@ pub const CancelKind = enum { user, auto };
 pub const AnyTask = struct {
     awaitable: Awaitable,
     coro: Coroutine,
-    state: std.atomic.Value(State),
```
lalinsky, line 160: @claude you can store the packed struct in atomic directly

src/task.zig
```diff
@@ -170,14 +170,45 @@ pub const AnyTask = struct {
...
+
+    pub inline fn storeState(self: *AnyTask, s: State, ordering: std.builtin.AtomicOrder) void {
+        self.state.store(s, ordering);
+    }
+
+    pub inline fn cmpxchgWeakState(self: *AnyTask, expected: State, desired: State, success_order: std.builtin.AtomicOrder, fail_order: std.builtin.AtomicOrder) ?State {
+        return self.state.cmpxchgWeak(expected, desired, success_order, fail_order);
+    }
+
+    pub inline fn cmpxchgStrongState(self: *AnyTask, expected: State, desired: State, success_order: std.builtin.AtomicOrder, fail_order: std.builtin.AtomicOrder) ?State {
```
lalinsky: @claude what is the value of these methods?

lalinsky: @claude remove them

lalinsky:
> 🧹 Nitpick comments (1)
> > src/runtime.zig (1)> `457-557`: **Please add a targeted regression test for wake-before-park.**
> > > This path is subtle and worth pinning with a stress/looped test that wakes in the tiny window before park and asserts no lost wake.
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against the current code and only fix it if needed.
> > > 
> > > In `@src/runtime.zig` around lines 457 - 557, Add a deterministic regression test
> > > that repeatedly exercises the "wake-before-park" window so we don't lose wakes:
> > > spawn an Executor and create a task (AnyTask) that yields/parks (so
> > > processCleanup -> .park path runs), then in a tight loop from another thread
> > > call the public wake API that triggers the state transitions (code paths
> > > involving Executor.scheduleTaskLocal/scheduleTaskRemote and the awaken bit on
> > > AnyTask.State) immediately before the task reaches park; repeat many iterations
> > > and assert the task is always rescheduled and completes (or increments a shared
> > > counter) rather than losing the wake. Ensure the test covers both same-executor
> > > and remote-executor cases (use getCurrentExecutorOrNull differences) and runs
> > > long enough to stress the cmpxchgWeak window in processCleanup/park and the wake
> > > handling at the top of the snippet so any regression is caught.
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> ℹ️ Review info

@coderabbitai do you have a suggestion for the stress test?

lalinsky:
@claude implement the stress test

PR #369, opened 2026-04-30
net: store handle in Stream.Reader/Writer, add fromStd constructors

lalinsky:
## Summary

- `Stream.Reader` and `Stream.Writer` now store a raw `Handle` instead of a full `Stream`/`Socket`, avoiding an unnecessary copy of the address
- Extracted `receiveBufImpl` and `sendBufImpl` free functions that operate directly on a `Handle`; `Socket.receiveBuf`/`sendBuf` delegate to these
- Removed `Stream.writeSplatHeader` (now internal as `sendSplatHeader`)
- Added `Stream.Reader.fromStd` and `Stream.Writer.fromStd` constructors that wrap a `std.Io.net.Stream` with timeout support, matching the stdlib `(stream, io, buffer)` signature

### DIFF
--- a/src/net.zig
+++ b/src/net.zig
@@ -947,9 +947,7 @@ pub const Socket = struct {
 
     /// Low-level receive function that accepts ev.ReadBuf directly.
     pub fn receiveBuf(self: Socket, buf: ev.ReadBuf, timeout: Timeout) !usize {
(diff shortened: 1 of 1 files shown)


lalinsky:
@claude disable tcpConnectToHost test on netbsd

PR #378, opened 2026-05-10
io: implement process spawn with non-blocking pipes on POSIX

lalinsky:
## Summary
- Implement `processSpawn` and `processSpawnPath` by delegating to `std.Io.Threaded` using the runtime's allocator
- Set child stdin/stdout/stderr pipes to non-blocking mode on POSIX so reads/writes integrate with the event loop
- Implement `processReplace` and `processReplacePath` by delegating to `globalIo`

## Test plan
- [x] `./check.sh --filter child --backend epoll` passes


lalinsky:
@coderabbitai review

PR #381, opened 2026-05-11
io: implement dirCreateDirPath and dirCreateDirPathOpen

lalinsky:
## Summary
- Implement `dirCreateDirPath` to create nested directory paths
- Implement `dirCreateDirPathOpen` to create and open nested paths
- Mark `processExecutableOpen` and `progressParentFile` as unsupported

## Details
The implementation follows std.Io.Threaded's algorithm: iterate from the target path back to existing ancestors, then forward creating each component.

## Test plan
- Added tests for both `createDirPath` and `createDirPathOpen`
- Tests verify nested directory creation and idempotent behavior

### DIFF
--- a/src/io.zig
+++ b/src/io.zig
@@ -2695,19 +2695,20 @@ test "io: dir createDirPath creates nested directories" {
     const io = rt.io();
 
     const dir: Io.Dir = .cwd();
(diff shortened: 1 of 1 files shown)


lalinsky:
@claude fix the windows test

PR #388, opened 2026-05-13
std.Io: implement concurrent batch operations

lalinsky:
## Summary
- Add `ev.DeviceIoControl` operation to make ioctl consistent with other ev operations
- Implement true concurrent batch operations via `awaitConcurrent`

## Details

Operations are now submitted to the event loop in parallel and completion callbacks move them to the completed list as they finish.

Key components:
- `BatchCompletionData`: tagged union holding ev operation + auxiliary data (iovecs, address storage, etc.)
- `BatchState`: memory pool for completion data, stored in `batch.userdata`
- Callbacks signal a waiter when operations complete
- Timeout support via `ev.Timer` racing with operations

The single-operation fast path is preserved - it calls `operateInner` directly without allocating from the pool.

## Test plan
- [x] `io: batch awaitConcurrent with empty batch returns immediately`
- [x] `io: batch awaitConcurrent with two net_receive operations`
- [x] All existing io tests pass


lalinsky:
@coderabbitai full review

PR #394, opened 2026-05-17
io: detect pollable fds for streaming read/write dispatch

lalinsky:
On epoll/kqueue backends, pipes/devices should be polled directly instead of going through the thread pool. This PR:

- Defaults `FileOpenFlags.nonblocking` and `FileCreateFlags.nonblocking` to `true`
- Removes the `supportsNonBlockingFileIo()` guard from backends
- After open/create, detects pollable fds via `lseek` (POSIX) / `GetFileType` (Windows)
- Dispatches streaming reads/writes to `PipeRead`/`PipeWrite` for pollable fds, `FileReadStreaming`/`FileWriteStreaming` for regular files
- Adds `Runtime.pollable_fds` map for caching foreign fds (stdin/stdout/stderr)
- Same dispatch logic in batch path

### DIFF
--- a/src/ev/backends/common.zig
+++ b/src/ev/backends/common.zig
@@ -233,6 +233,10 @@ pub fn fileOpenWork(work: *Work) void {
     const file_open: *FileOpen = @fieldParentPtr("internal", internal);
     const loop = internal.linked_context.loop;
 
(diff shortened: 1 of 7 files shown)


lalinsky:
@claude extract the gist of this change, logical, not in code, to serve as a spec for rewriting this later

PR #396, opened 2026-05-19
Add allow_task_migration runtime option, enable by default

lalinsky:
## Summary

- Replaces the stub `canMigrate()` method on `AnyTask` (which always returned `false`) with an `allow_task_migration: bool` field on `RuntimeOptions`, defaulting to `true`
- Partially reverts #393 (simplify current task/executor tracking), which introduced a hang in release mode when task migration was enabled; keeps the migration-safe `c.loop`-based cancellation fixes from that PR
- Migration-specific tests now explicitly set `allow_task_migration = true` so they exercise migration regardless of the default


lalinsky:
@claude analyze why is the recently added test failing after task migration

PR #397, opened 2026-05-19
Add deterministic test for RwLock last-reader/writer wakeup race

manishrjain:
## Summary

The deadlock fixed in #395 (signal → broadcast in `unlockShared`) is reproducible deterministically. This PR adds a regression test for it.

I had hit the same deadlock in a downstream project and wrote my own fix + test before noticing #395 already landed upstream with the same one-line fix. Rather than landing a duplicate fix, this PR carries forward just the test — upstream currently has no reproducer for this specific race.

## How the test works

The race needs a specific queue shape: a reader is parked on the shared cond *before* the writer that will need the final reader's wake. Without `broadcast`, the queued reader consumes the signal, sees `writers_waiting > 0`, goes back to sleep, and the writer is stranded.

The test constructs that shape by temporarily bumping `writers_waiting` while spawning a reader (so the reader parks), then spawning the real writer. It then asserts the writer wakes when the last reader releases.

If the broadcast fix is reverted, this test deadlocks; the test then recovers via a manual `broadcast` so the runtime can `deinit` cleanly, and the assertion fails. (Verified locally by reverting #395 and running the test.)

## Test plan

- [x] Passes against current `main` (post-#395)
- [x] Fails (assertion, not hang — recovery path engages) if `broadcast` is reverted to `signal`

### DIFF
--- a/src/sync/RwLock.zig
+++ b/src/sync/RwLock.zig
@@ -206,6 +206,83 @@ test "RwLock basic shared lock/unlock" {
     rwlock.unlock();
 }
 
+test "RwLock last reader wakes writer when reader is queued first" {
+    const runtime = try Runtime.init(std.testing.allocator, .{ .executors = .exact(4) });
+    defer runtime.deinit();
+
+    var rwlock = RwLock.init;
+    var writer_acquired = std.atomic.Value(bool).init(false);
+
+    try rwlock.lockShared(); // Read Lock A
+
+    const TestFn = struct {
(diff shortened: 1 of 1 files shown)


manishrjain:
Just realized it's a duplicate of https://github.com/lalinsky/zio/pull/395 . I was on the v0.11 branch and had to learn about this bug the hard way.

lalinsky:
Thank you for the test!

PR #411, opened 2026-05-26
When resolv.conf is missing, fallback to default

psznm:
When resolv.conf is missing, resolver should use default 127.0.0.1:53


lalinsky (approved):
Looks good, thank you.

PR #414, opened 2026-05-27
dns: extend CacheKey to handle names longer than 31 characters

lalinsky:
The previous `CacheKey` silently skipped caching and deduplication for
hostnames longer than 31 characters. Long names (AWS endpoints, generated
subdomains) are common in practice and benefit from both.

## New layout (48 bytes)

```
hash: u64    — Wyhash of the full name, used for slot index and equality
len:  u8     — full name length, panics on overflow (DNS max is 253)
buf:  [39]u8 — first 39 bytes of the name, zero-padded
```

Collision requires matching `len`, `hash`, and the first 39 chars
simultaneously — negligible in practice (~1/2^64 per pair).

`CacheKey.init` now always returns `CacheKey` (non-optional), removing
the length guard from all callers.

### DIFF
--- a/src/dns/resolver/cache.zig
+++ b/src/dns/resolver/cache.zig
@@ -6,28 +6,28 @@ const net = @import("../../net.zig");
 const Timestamp = @import("../../time.zig").Timestamp;
 
 pub const max_cached_addrs = 3;
-pub const max_cached_name_len = 31;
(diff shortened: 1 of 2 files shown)


lalinsky:
@coderabbitai review

PR #429, opened 2026-05-28
Simplify BroadcastChannel

antlanc7:
This PR simplifies how `BroadcastChannel` consumers are created and managed.

Before this change, callers had to create a `Consumer`, pass it to `subscribe()`, and remember to call `unsubscribe()` afterwards. 
I noticed this was necessary just because the channel did keep track of active consumers by pointer, but actually the list was unused, so `subscribe()` can just return a `Consumer` directly and we can get rid of the tracking list.

So usage goes from:

```zig
var consumer = BroadcastChannel(u32).Consumer{};
channel.subscribe(&consumer);
defer channel.unsubscribe(&consumer);
```

to:

```zig
var consumer = channel.subscribe();
```

The internal consumer list and the `unsubscribe()` method have been removed, and the tests have been updated to use the new API.

This is my first contribution here, thank you for the library, it's awesome!

### DIFF
--- a/src/sync/broadcast_channel.zig
+++ b/src/sync/broadcast_channel.zig
@@ -274,9 +274,7 @@ const AsyncReceiveImpl = struct {
 /// }
 ///
 /// fn listener(rt: *Runtime, ch: *BroadcastChannel(u32)) !void {
-///     var consumer = BroadcastChannel(u32).Consumer{};
-///     ch.subscribe(&consumer);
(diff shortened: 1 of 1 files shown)


lalinsky:
Thank you, this is good. The consumer list was there because I originally wanted to support a mode where slow consumers would block the publisher, but never got to implementing it. It's probably not useful anyway. So it can go away.

Issue #262, opened 2026-01-30
Optimistically leave socket fd in epoll/kqueue


lalinsky:
When doing a read, we could possibly leave the socket fd in the epoll/kqueue system-level queue, so that when the next read request comes, we already knows the readiness. It would require readiness tracking per fd, but I think we need some kind of fd metadata map in ev.Loop anyway. This doesn't translate to io_uring, because that would require another buffer to read into.

lalinsky:
Fixed https://github.com/lalinsky/zio/pull/526

PR #443, opened 2026-05-31
ev/kqueue: fix completion-lifecycle use-after-free under concurrent cancellation

### LINKED ISSUE #442: kqueue backend: use-after-free under concurrent cancellation
## Summary

On the kqueue backend (macOS/BSD), `poll()` can dereference a `Completion` that has already been freed, crashing the process. It reproduces when many fibers blocked on socket I/O are cancelled concurrently from sibling fibers on other executors (e.g. listener / endpoint teardown).

## Root cause

`submit` stores a raw completion pointer in the kevent cookie (`udata = @intFromPtr(completion)`) and `poll()` recovers it with `@ptrFromInt(event.udata)`. But a `Completion` lives on its fiber's stack and is freed the moment that fiber unwinds — there is no free-on-completion contract like io_uring's one-CQE-per-SQE.

So within a single `poll()` batch, completing `event[i]` resumes a fiber that cancels a sibling fiber, freeing that sibling's `Completion`, while a stale / `EV_EOF` `event[j]` later in the same batch still carries its pointer. `poll()` then reads `completion.state` out of freed memory.

Neither sibling backend has this problem:

- **epoll** keys by the fd (`epoll_event.data.fd`), re-derives the completion from a loop-owned `poll_queue`, and `EPOLL_CTL_DEL` is synchronous.
- **io_uring** is a completion model (free-on-CQE), so its raw `user_data` is never stale.

kqueue is the one reactor backend that stores a raw pointer with neither safeguard.

## Reproduction

8 fibers, each blocked in `receiveMany` on its own datagram socket, cancelled concurrently from sibling fibers on a 2-executor runtime. On `main` this crashes (`EXC_BAD_ACCESS` in `kqueue.poll` while reading `completion.state`) on a large fraction of runs. A minimal repro is included as a new `src/io.zig` test in the accompanying PR.

```
frame #0: ev.backends.kqueue.poll ... at kqueue.zig  (ldrb of completion.state)
frame #1: ev.loop.Loop.tick
frame #2: runtime.Executor.run
```

## Environment

zig 0.16.0, zio `main`, macOS arm64.

## Fix

A PR follows that makes the kqueue backend mirror epoll (key knotes by `(ident, filter)` in a loop-owned `poll_queue`, re-derive the completion from that table in `poll()`, and apply changes synchronously).

GrapeBaBa:
## Problem

The kqueue backend stores a raw `*Completion` in the kevent `udata` and recovers it via `@ptrFromInt` in `poll()`. Because a `Completion` is freed when its fiber unwinds, completing `event[i]` in a poll batch can resume a fiber that cancels a sibling and frees that sibling's completion, while a stale `event[j]` in the same batch still points at it — `poll()` then dereferences freed memory (`EXC_BAD_ACCESS` reading `completion.state`). It surfaces under concurrent cross-executor cancellation (listener / endpoint teardown).

Neither sibling backend has this: epoll re-derives the completion from a loop-owned `poll_queue` keyed by fd (the cookie is the fd, not a pointer); io_uring frees only on the op's own CQE.

## Fix

Mirror the epoll backend:

- Key each knote by `(ident, filter)` in a loop-owned `poll_queue` (`AutoHashMapUnmanaged`); `udata` is no longer used for routing. `poll()` re-derives the completion(s) via `poll_queue.get(...)`, never from the event cookie — a stale event resolves to a removed entry (skip) or a live completion (a real I/O retry), never freed memory.
- Knotes stay persistent level-triggered (no `EV_ONESHOT`), and the changelist is still **batched** into `poll()`'s `kevent()` as before — so a registration error still comes back as an `EV_ERROR` event handled by `checkCompletion` (preserves the existing pipe/socket error semantics).
- `EV.DELETE` changes carry a sentinel `udata`, so their `EV_ERROR` *receipts* (e.g. `ENOENT` when the knote was already gone — common after `poll`+`write` on one fd) are skipped instead of being mis-routed to a completion that recycled the same `(ident, filter)`.
- The `(ident, filter)` key is packed wide enough for NetBSD's `i32` `kevent.filter` (it is `i16` on Darwin/FreeBSD).

No public API change. `checkCompletion` / `handleKqueueError` (the per-op I/O logic) are untouched.

## Testing

- New `src/io.zig` test (`io: concurrent cross-executor cancel of N blocked recvmsg fibers is UAF-free`): 8 fibers blocked in `receiveMany`, cancelled concurrently on a 2-executor runtime. On `main` this crashes; with the fix it passes 25–30/30 under stress on macOS arm64.
- Full `zig build test` green on macOS arm64 (the only failure is the pre-existing, network-dependent `dns: lookup www.github.com follows CNAME`). Cross-compiles clean for `x86_64-freebsd` / `x86_64-netbsd`.

zig 0.16.0.

Fixes #442

### DIFF
--- a/src/ev/backends/kqueue.zig
+++ b/src/ev/backends/kqueue.zig
@@ -63,12 +63,23 @@ const EVFILT_USER: i16 = switch (builtin.target.os.tag) {
 };
 const NOTE_TRIGGER: u32 = 0x01000000;
 
+const delete_marker: usize = std.math.maxInt(usize);
+
+const PollEntry = struct {
+    completions: Queue(Completion),
+};
+
 allocator: std.mem.Allocator,
 kqueue_fd: i32 = -1,
 waker_ident: usize = undefined,
+poll_queue: std.AutoHashMapUnmanaged(u64, PollEntry) = .empty,
 change_buffer: std.ArrayList(std.c.Kevent) = .empty,
 events: []std.c.Kevent,
-queue_size: u16,
+
+fn makeKey(ident: usize, filter: i32) u64 {
+    std.debug.assert(ident <= std.math.maxInt(u32));
+    return (@as(u64, @intCast(ident)) << 32) | @as(u32, @bitCast(filter));
+}
 
 pub fn init(self: *Self, allocator: std.mem.Allocator, queue_size: u16, shared_state: *SharedState) !void {
     _ = shared_state;
@@ -109,11 +120,13 @@ pub fn init(self: *Self, allocator: std.mem.Allocator, queue_size: u16, shared_s
         .waker_ident = waker_ident,
         .change_buffer = change_buffer,
         .events = events,
-        .queue_size = queue_size,
     };
+
+    try self.poll_queue.ensureTotalCapacity(self.allocator, queue_size);
 }
 
 pub fn deinit(self: *Self) void {
+    self.poll_queue.deinit(self.allocator);
     self.change_buffer.deinit(self.allocator);
     self.allocator.free(self.events);
     if (self.kqueue_fd != -1) {
@@ -123,7 +136,6 @@ pub fn deinit(self: *Self) void {
 
 pub fn wake(self: *Self, state: *LoopState) void {
     _ = state;
-    // TODO: Could also submit other pending changes if we have any
     var changes: [1]std.c.Kevent = .{.{
         .ident = self.waker_ident,
         .filter = EVFILT_USER,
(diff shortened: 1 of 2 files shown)


lalinsky:
Hm, I'll have to think about this, because canceling a task should always cancel the operation, and actually wait for it to get canceled. I think the problem is that we call `EV_DELETE`, which cancels it immediately, but kqueue already has the completion events queued up for delivery, so they still arrive later. Maybe the correct approach would be to drain the queue first. Or maybe the map is indeed the best option. Or maybe the map should contains completion pointers to ignore in the next event drain. I'm not sure yet.

lalinsky:
Ok, given that I wanted to do https://github.com/lalinsky/zio/issues/262 eventually, the map is fine. I was experimenting with an alternative fix for this, but if we want to support multiple reads per fd, we would still need the map.

PR #445, opened 2026-05-31
os/time: drive clock_gettime through the vDSO without libc

zoptia:
now() reads the monotonic clock once per event-loop tick via std.posix.system.clock_gettime (= std.os.linux.clock_gettime). That serves the read from the kernel vDSO (userspace, no syscall) only once std.os.linux.elf_aux_maybe points at the ELF aux vector. Zig populates it from its own _start, but zio links libc (build.zig), so glibc owns _start and leaves elf_aux_maybe null — the vDSO lookup then fails and every clock read falls back to a real clock_gettime syscall. Under a high wakeup rate the process burns most of its CPU in that syscall.

Recover the aux vector from /proc/self/auxv (pure syscalls, no libc, no reliance on Zig owning _start) on first use and hand it to elf_aux_maybe; arch-specific vDSO symbol resolution is left to std's existing lookup, so x86_64 and aarch64 are both covered. The one-time fill is guarded by a state machine so concurrent now() callers synchronize before clock_gettime runs its one-shot lookup. Best effort: if /proc is unavailable the read stays on the syscall path.

Adds a regression test: because the test binary links libc, elf_aux_maybe is null without the fix and the test fails.

Measured on a libc-linked QUIC download client: clock_gettime syscalls per 32 MB transfer dropped from ~43k to 0; loopback throughput +8% (GSO sender) to +26% (paced sender).

### DIFF
--- a/src/os/time.zig
+++ b/src/os/time.zig
@@ -48,6 +48,15 @@ pub fn now(clock: Clock) Timestamp {
                 .monotonic => posix.system.CLOCK.MONOTONIC,
                 .realtime => posix.system.CLOCK.REALTIME,
             };
+            // std.posix.system.clock_gettime serves the read from the
+            // kernel vDSO (userspace, no syscall) — but only once
+            // std.os.linux.elf_aux_maybe points at the ELF aux vector.
+            // Zig sets that from its own _start; when the program links
+            // libc (zio does), glibc owns _start and leaves it null, so
+            // every read falls back to a real clock_gettime syscall.
+            // now() is the event loop's hottest call, so recover the aux
+            // vector ourselves (pure syscalls, no libc) on first use.
+            if (builtin.os.tag == .linux) ensureVdsoAux();
             // TODO: use our posix layer, not std.posix
             // https://codeberg.org/ziglang/zig/pulls/35506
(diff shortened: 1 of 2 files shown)


lalinsky:
Damn it, this is confusing. Thank you for doing this workaround. I'll need to do a bit of a research before merging.

lalinsky:
@zoptia The thing I'm confused about:

When linking libc, `std.posix.system` points to `std.c` even on Linux, so I made https://github.com/lalinsky/zio/commit/b241755bf59d97523b4a489f0316b85aec9a6207 to work around the issue

So now it should use the libc `clock_gettime` which should use vDSO internally.

Doesn't this change we need to revert https://github.com/lalinsky/zio/commit/b241755bf59d97523b4a489f0316b85aec9a6207 ?

lalinsky:
Plus, what I would prefer is something similar https://codeberg.org/ziglang/zig/pulls/35506, that means using the libc function to get the value. But maybe just using the libc `clock_gettime` is fine as well. I'm getting no clock syscalls when using zio master.

lalinsky:
I'm going to close this, according to https://codeberg.org/ziglang/zig/pulls/35506 we should use the C function, which we now do, so it should use vDSO through there.

PR #448, opened 2026-06-01
Use splat instead of ** and *Sentinel instead of *Z

psznm:
Backwards compatible changes required for upgrade to zig 0.17


lalinsky:
Thank you!

PR #452, opened 2026-06-02
fs: fix macOS dirDeleteFile returning AccessDenied instead of IsDir f…

sdzx-1:
…or directories

On macOS, unlinkat() returns EPERM (not EISDIR) when the target is a directory and AT_REMOVEDIR is not set.  The errno-to-error mapping was treating EPERM as error.AccessDenied, giving a misleading error when dirDeleteFile was called on a directory.

Add a Darwin-specific fstatat() check in the EPERM path to detect directories and return error.IsDir instead.  Other EPERM causes (immutable flag, genuine permission issues) still return AccessDenied.

Fixes #

### DIFF
--- a/src/os/fs.zig
+++ b/src/os/fs.zig
@@ -1137,6 +1137,24 @@ pub fn dirDeleteFile(allocator: std.mem.Allocator, dir: fd_t, path: []const u8) 
         switch (posix.errno(rc)) {
             .SUCCESS => return,
             .INTR => continue,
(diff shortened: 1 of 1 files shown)


sdzx-1:
https://github.com/getz3/z3/actions/runs/26793771633/job/78985671289#step:14:48

The `deleteTree` uses `dirDeleteFile`, and I encountered this error during testing on macOS.

PR #455, opened 2026-06-02
fs: add resolve_beneath support

### LINKED ISSUE #385: Add support for resolve_beneath=true
- [x] Linux
- [x] macOS
- [x] FreeBSD
- [ ] Windows
- [ ] Fallback for other systems

lalinsky:
## Summary

- **Linux 5.6+**: uses `openat2(RESOLVE_BENEATH | NO_MAGICLINKS)` natively — via `IORING_OP_OPENAT2` in the io_uring backend or direct syscall in epoll/blocking backends; falls back with a warning on older kernels
- **FreeBSD 13+ / macOS**: uses `O_RESOLVE_BENEATH` flag in `openat` (both return `ENOTCAPABLE` on escape)
- **Other platforms**: logs a warning, no enforcement
- `EXDEV` (Linux openat2) and `ENOTCAPABLE` (FreeBSD/Darwin) are mapped to the new `error.PathEscaped` in `FileOpenError` and `DirOpenError`
- At the `std.Io` layer, `PathEscaped` maps to `error.AccessDenied` (matching the stdlib convention)

This goes further than `std.Io.Threaded`, which only handles FreeBSD/macOS via `O_RESOLVE_BENEATH` and has no Linux openat2 support.

Fixes https://github.com/lalinsky/zio/issues/385

## Test plan

- [ ] Test passes on Linux (io_uring and epoll backends)
- [ ] Check CI results for macOS and FreeBSD
- [ ] Check CI results for Windows and other platforms (should warn and not crash)

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -13,9 +13,17 @@ pub fn build(b: *std.Build) void {
         "Override the default event loop backend (io_uring, epoll, kqueue, iocp, poll)",
     );
 
+    const ResolveBeneathMode = enum { strict, best_effort };
+    const resolve_beneath_mode = b.option(
+        ResolveBeneathMode,
+        "resolve-beneath-mode",
+        "How to handle resolve_beneath on platforms without kernel support: strict (error.Unsupported) or best_effort (log warning, continue)",
+    ) orelse .strict;
+
     // Create options for backend selection
     var options = b.addOptions();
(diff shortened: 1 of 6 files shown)


lalinsky:
@coderabbitai review

PR #469, opened 2026-06-06
Add file-to-socket sendfile (ev.NetSendFile + native Stream.Writer.sendFile)

lalinsky:
## What

Adds file-to-socket transfer ("sendfile") to zio, surfaced through the native `net.Stream.Writer` as the `std.Io.Writer` `sendFile`/`sendFileAll` methods.

## Why this lives in the native writer

`std.Io` 0.16 leaves the optimized file→socket path entirely unwired: `net.Stream.Writer.sendFile` is a `return error.Unimplemented` stub that never dispatches to the `netWriteFile` vtable slot, and every backend's `netWriteFile` is a TODO/panic. So filling zio's `netWriteFile` vtable would be unreachable code. Instead, zio's **own** `net.Stream.Writer` implements the `Io.Writer` `sendFile` vtable method, which `Io.Writer.sendFile`/`sendFileAll` *do* dispatch to — making this the only reachable surface. `io.zig`'s `netWriteFile` stays unimplemented.

## How

**`ev.NetSendFile`** — a new event-loop operation transferring a file to a socket:
- New `net_send_file` capability + `Op`. No backend advertises it yet, so the loop drives a generic **read/write callback loop** (`Fallback`): read a chunk from the file into a 16 KiB buffer, send it to the socket, repeat until EOF or `limit`. Socket short-writes drain the buffer before the next read. Exactly one child op (`FileRead`/`NetSend`) is in flight at a time; each completion re-arms the next.
- Cancellation is forwarded to whichever child is in flight.
- The no-runtime (`executeBlocking`) path runs the same loop synchronously.
- `.net_send_file => unreachable` arms added to every backend's exhaustive op switches; the op never reaches a backend.

**Native `Stream.Writer.sendFile`** — implements the `std.Io.Writer` vtable method: flush the writer's buffered bytes (the "header") and the reader's buffered bytes first, then stream the file body from `file_reader.pos` via `ev.NetSendFile`. `limit` caps only file-origin bytes; the return value is the file bytes (header excluded via `consume`), and EOF is signaled with `error.EndOfStream` — matching the `File.Writer` reference impl and the `sendFileAll` contract.

An io_uring `splice` fast path can be added later behind the same capability flag and API.

## Testing

- `src/ev/test/sendfile.zig`: multi-iteration streaming (40 KB), `limit` enforcement, and the blocking no-runtime path.
- `src/net.zig`: end-to-end `Stream.Writer.sendFileAll` over loopback TCP (40 KB, every byte verified).

Verified green on io_uring / epoll / poll (full suites), Windows via wine, and kqueue/macOS compiles.


lalinsky:
@claude analyze the option of using double buffering, so that we can do the subsequent read and write concurrentl

lalinsky:
@coderabbitai create two follow up issues, one for specialized io_uring version and one for IOCP

src/ev/test/sendfile.zig
```diff
@@ -0,0 +1,192 @@
...
+        };
+        self.total += n;
+        if (n != 0 and self.total < self.buf.len) {
+            self.start(loop); // re-arm for the remainder
+        }
+    }
+};
+
+test "sendfile: NetSendFile sends a file over a socket (fallback)" {
+    if (builtin.os.tag == .windows) return error.SkipZigTest;
```
lalinsky: @coderabbitai if we use TCP socket pairs, these tests should be good on windows, right?

PR #474, opened 2026-06-06
Implement lockStderr/unlockStderr and expose debug_io for free logFn

### LINKED ISSUE #337: Async logging

lalinsky:
Closes #337

Implements the `std.Io` vtable's `lockStderr`, `tryLockStderr`, and `unlockStderr` using a module-global zio `Mutex` and a lazily-initialized `std.Io.File.Writer` for stderr.

With this in place, users get async-aware logging for free by setting:

```zig
pub const std_options_debug_io: std.Io = @import("zio").debug_io;
```

`std.log.defaultLog` already calls `lockStderr`/`unlockStderr`, so no custom `logFn` is needed. Inside a zio task the mutex acquisition and stderr write cooperatively yield; outside a task they fall back to blocking syscalls on the OS thread.

`debug_io` is a comptime constant (`userdata = null`) — all write-path vtable functions use TLS task detection (`waitForIo`) rather than the runtime pointer, so no runtime userdata is needed.

Adds 4 tests: blocking fallback, async path, concurrent tasks, and `std.log.defaultLog` integration. All 485 tests pass.

### DIFF
--- a/src/io.zig
+++ b/src/io.zig
@@ -57,6 +57,19 @@ pub fn fromRuntime(rt: *Runtime) Io {
     };
 }
 
+/// A `std.Io` value that routes through zio's vtable with no runtime userdata.
+///
+/// All I/O operations use TLS task detection (see `waitForIo`): inside a
+/// zio task they cooperatively yield; outside a task they fall back to
+/// blocking syscalls.  Operations that truly need a `*Runtime` (spawn,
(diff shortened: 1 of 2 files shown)


lalinsky:
@claude should we use some buffer for the writer? How does std.io.threaded initialize it?

PR #511, opened 2026-06-23
ev/epoll: keep sockets registered via a shared cross-loop registration table

### LINKED ISSUE #510: epoll backend: ~4 epoll_ctl per request (arm/disarm) caps throughput ~25-35% below io_uring
While benchmarking a plain TCP echo server I found the `epoll` backend is noticeably slower than the `io_uring` backend. The root cause is that epoll interest is armed and disarmed for every single operation, costing ~4 `epoll_ctl` per echo round-trip.

## Setup

8-core EPYC 7402, Arch Linux (kernel 7.0), Zig 0.16.0. Closed-loop TCP ping-pong (depth 1), 64-byte messages, system-default sockets (no `TCP_NODELAY`). Server pinned with `taskset` so it is the bottleneck; the same client drives every server, 10s per run. Echo loop is `read -> write` with no artificial delay, written against the `std.Io` interface (the `tcp_echo_server.zig` example has a `sleep(1000ms)`, so I used a stripped-down server).

## Results (req/s)

Single server thread (pinned to 1 core):

| conns | epoll before | epoll after | io_uring | asio (1 thread) |
|------:|-------------:|------------:|---------:|----------------:|
| 50    | 93.8k        | **121.5k**  | 126.5k   | 125.6k |
| 200   | 81.7k        | **105.6k**  | 109.9k   | 126.9k |
| 1000  | 74.5k        | **87.0k**   | 88.9k    | 97.0k |

3 server threads (pinned to 3 cores):

| conns | epoll before | epoll after | io_uring |
|------:|-------------:|------------:|---------:|
| 50    | 266.5k       | **368.6k**  | 373.8k |
| 200   | 276.1k       | **380.8k**  | 390.2k |
| 1000  | 263.3k       | **378.9k**  | 379.8k |

Before the fix epoll is ~25-35% below io_uring. After the fix it lands within a few percent of io_uring everywhere. (asio is just a single-thread sanity reference; the apples-to-apples comparison is epoll vs io_uring within zio.)

## Root cause

`strace -f -c` on the single-thread server, per request:

```
epoll backend:  epoll_ctl 4.02/req   recvfrom 1.00   sendto 1.00   epoll_pwait2 0.20
```

The backend is a one-shot, level-triggered reactor: `submit()` calls `addToPollQueue()` -> `epoll_ctl(ADD)`, and once the fd fires, `removeFromPollQueue()` -> `epoll_ctl(DEL)`. One echo round (recv + send) is ADD/DEL twice = ~4 `epoll_ctl`, which is ~55% of kernel time. Worth noting `net_connect` already avoids this (it tries `connect()` first and only registers on `WouldBlock`); `recv`/`send`/`accept` don't.

## Fix

Two parts:

1. **Speculative syscall.** In `submit()`, try `recv`/`send`/`accept` (and the `*from`/`*msg` variants) directly first; only `addToPollQueue()` on `WouldBlock`. Same shape as the existing `net_connect` handler. This removes the arm/disarm for sends and for any already-ready case.

2. **Edge-triggered + persistent registration.** Register each fd once with `EPOLLIN|EPOLLOUT|EPOLLET` and keep it for the fd's lifetime; `removeFromPollQueue()` no longer `DEL`s, teardown moves to the `net_close` handler. The mask is constant so there is no `MOD` either. `EPOLLET` is safe here precisely because (1) guarantees we always drain to `EAGAIN` before we ever wait in epoll, so no edge is lost.

(1) on its own isn't enough: on a fast server the next `recv` almost always `EAGAIN`s (the peer's data hasn't arrived yet) and re-arms, so (2) is what removes the recv-side churn. After both, `strace` shows `epoll_ctl` ~0/req (just the lifetime ADD/DEL per fd).

## Validation

With `-Dbackend=epoll`: the `net`, `Stream`, `recv`, `send`, `accept`, `cancel`, and `timeout` test filters all pass. The DNS tests (which exercise the UDP `recvfrom`/`sendto` path) are 35/37 — the two failures are pre-existing hardcoded-IP/CNAME assertions against live DNS, unrelated to this change. All benchmark runs report 0 errors.

## Notes on the patch

- I arm `IN|OUT|ET` unconditionally, so there's one spurious `EPOLLOUT` wakeup per socket at registration time (negligible). Could be narrowed to arm `OUT` only on a send `EAGAIN` if you'd prefer.
- Teardown is done in `net_close`; the kernel also drops a closed fd from epoll automatically, the explicit `DEL` just keeps `poll_queue` clean for fd reuse.
- Dropped the `entry.type` debug assert, since a socket now keeps a single `poll_queue` entry across `connect -> recv/send`.

Happy to turn this into a PR if the approach looks reasonable.

<details>
<summary>Full benchmark output (raw, with latency percentiles)</summary>

```
== single server thread (cpu0), client cpu1-7 x6, 10s, 64B ==
epoll-before  conns=50   req=937633  rps=93763  mbps=11.4 avg_us=532.7 p50=528   p99=638   p999=1285  max=17020
epoll-after   conns=50   req=1215454 rps=121545 mbps=14.8 avg_us=410.8 p50=376   p99=783   p999=942   max=2540
io_uring      conns=50   req=1264506 rps=126451 mbps=15.4 avg_us=394.8 p50=392   p99=450   p999=771   max=2400
asio          conns=50   req=1255515 rps=125552 mbps=15.3 avg_us=397.7 p50=396   p99=442   p999=680   max=2281

epoll-before  conns=200  req=816606  rps=81661  mbps=10.0 avg_us=2450.6 p50=2470 p99=3245  p999=4406  max=36377
epoll-after   conns=200  req=1055603 rps=105560 mbps=12.9 avg_us=1895.5 p50=1877 p99=2226  p999=3026  max=30807
io_uring      conns=200  req=1098522 rps=109852 mbps=13.4 avg_us=1821.0 p50=1825 p99=1995  p999=3123  max=43214
asio          conns=200  req=1269034 rps=126903 mbps=15.5 avg_us=1576.7 p50=1523 p99=2076  p999=3342  max=30736

epoll-before  conns=1000 req=744745  rps=74474  mbps=9.1  avg_us=13676  p50=13350 p99=15256 p999=200000 max=200000
epoll-after   conns=1000 req=869989  rps=86999  mbps=10.6 avg_us=11713  p50=11419 p99=14090 p999=200000 max=200000
io_uring      conns=1000 req=888945  rps=88894  mbps=10.9 avg_us=11462  p50=11246 p99=13396 p999=200000 max=200000
asio          conns=1000 req=970065  rps=97006  mbps=11.8 avg_us=10188  p50=10242 p99=11673 p999=14374  max=200000

== 3 server threads (cpu0-2), client cpu3-7 x5, 10s, 64B ==
epoll-before  conns=50   req=2665155 rps=266516 mbps=32.5 avg_us=187.0 p50=178  p99=563  p999=1010 max=6033
epoll-after   conns=50   req=3686174 rps=368617 mbps=45.0 avg_us=135.0 p50=111  p99=564  p999=989  max=5940
io_uring      conns=50   req=3737879 rps=373788 mbps=45.6 avg_us=133.1 p50=127  p99=556  p999=956  max=4409

epoll-before  conns=200  req=2761167 rps=276117 mbps=33.7 avg_us=724.2 p50=705  p99=1297 p999=1948 max=28101
epoll-after   conns=200  req=3808478 rps=380848 mbps=46.5 avg_us=524.9 p50=477  p99=1271 p999=1965 max=16755
io_uring      conns=200  req=3902315 rps=390232 mbps=47.6 avg_us=512.3 p50=502  p99=1151 p999=1873 max=55467

epoll-before  conns=1000 req=2633160 rps=263316 mbps=32.1 avg_us=3862.1 p50=3770 p99=5074 p999=7293 max=200000
epoll-after   conns=1000 req=3789067 rps=378907 mbps=46.3 avg_us=2687.8 p50=2626 p99=3872 p999=6280 max=200000
io_uring      conns=1000 req=3798178 rps=379818 mbps=46.4 avg_us=2673.2 p50=2623 p99=3943 p999=7058 max=200000
```

</details>

<details>
<summary>epoll.zig diff</summary>

```diff
diff --git a/src/ev/backends/epoll.zig b/src/ev/backends/epoll.zig
index 6daccfc..ebbf5d2 100644
--- a/src/ev/backends/epoll.zig
+++ b/src/ev/backends/epoll.zig
@@ -21,6 +21,7 @@ const NetSendMsg = @import("../completion.zig").NetSendMsg;
 const NetPoll = @import("../completion.zig").NetPoll;
 const PipePoll = @import("../completion.zig").PipePoll;
 const PipeClose = @import("../completion.zig").PipeClose;
+const NetClose = @import("../completion.zig").NetClose;
 const ProcessWait = @import("../completion.zig").ProcessWait;
 const fs = @import("../../os/fs.zig");
 const linux = std.os.linux;
@@ -201,12 +202,18 @@ fn addToPollQueue(self: *Self, state: *LoopState, fd: NetHandle, completion: *Co
     };
 
     var entry = gop.value_ptr;
-    const op_events = getEvents(completion);
 
     if (!gop.found_existing) {
+        // Register the fd once, edge-triggered, for the rest of its lifetime.
+        // Both IN and OUT are armed so we never need epoll_ctl(MOD) when the
+        // socket switches between reading and writing, and we never disarm on
+        // completion (see removeFromPollQueue). Edge-triggered means a ready fd
+        // with no waiter won't spin; correctness relies on submit() always
+        // attempting the syscall first (draining to EAGAIN) before it waits.
+        const persist_events = std.os.linux.EPOLL.IN | std.os.linux.EPOLL.OUT | std.os.linux.EPOLL.ET;
         var event = std.os.linux.epoll_event{
             .data = .{ .fd = fd },
-            .events = op_events,
+            .events = persist_events,
         };
         const rc = std.os.linux.epoll_ctl(self.epoll_fd, std.os.linux.EPOLL.CTL_ADD, fd, &event);
         switch (posix.errno(rc)) {
@@ -222,81 +229,26 @@ fn addToPollQueue(self: *Self, state: *LoopState, fd: NetHandle, completion: *Co
         entry.* = .{
             .completions = .{},
             .type = getPollType(completion.op),
-            .events = op_events,
+            .events = persist_events,
         };
         entry.completions.push(completion);
         return;
     }
 
-    std.debug.assert(entry.type == getPollType(completion.op));
-
-    const new_events = entry.events | op_events;
-    if (new_events != entry.events) {
-        var event = std.os.linux.epoll_event{
-            .events = new_events,
-            .data = .{ .fd = fd },
-        };
-        const rc = std.os.linux.epoll_ctl(self.epoll_fd, std.os.linux.EPOLL.CTL_MOD, fd, &event);
-        switch (posix.errno(rc)) {
-            .SUCCESS => {},
-            else => |err| {
-                log.err("Failed to epoll_ctl(CTL_MOD): {}", .{err});
-                completion.setError(error.Unexpected);
-                state.markCompletedFromBackend(completion);
-                return;
-            },
-        }
-        entry.events = new_events;
-    }
+    // fd already registered (persistent, constant mask): just attach the
+    // completion. No epoll_ctl needed. The entry type may legitimately differ
+    // from the first op (e.g. connect -> recv/send on the same socket), so we
+    // no longer assert on it.
     entry.completions.push(completion);
 }
 
 fn removeFromPollQueue(self: *Self, fd: NetHandle, completion: *Completion) !void {
     const entry = self.poll_queue.getPtr(fd) orelse return;
-
     _ = entry.completions.remove(completion);
-
-    if (entry.completions.head == null) {
-        // No more completions - remove from epoll and poll queue
-        const del_rc = std.os.linux.epoll_ctl(self.epoll_fd, std.os.linux.EPOLL.CTL_DEL, fd, null);
-        const err = posix.errno(del_rc);
-
-        // Always remove from poll_queue when list is empty to avoid stale entries
-        // (fd will be auto-removed from epoll when closed anyway)
-        const was_removed = self.poll_queue.remove(fd);
-        std.debug.assert(was_removed);
-
-        switch (err) {
-            .SUCCESS, .NOENT, .BADF => {
-                // SUCCESS: successfully removed
-                // NOENT: fd was not registered (already removed or never added) - safe to proceed
-                // BADF: fd was closed (and auto-removed from epoll) - safe to proceed
-            },
-            else => return unexpectedError(err),
-        }
-        return;
-    }
-
-    // Recalculate events from remaining completions
-    var new_events: u32 = 0;
-    var iter: ?*Completion = entry.completions.head;
-    while (iter) |c| : (iter = c.next) {
-        new_events |= getEvents(c);
-    }
-
-    if (new_events != entry.events) {
-        var event = std.os.linux.epoll_event{
-            .events = new_events,
-            .data = .{ .fd = fd },
-        };
-        const mod_rc = std.os.linux.epoll_ctl(self.epoll_fd, std.os.linux.EPOLL.CTL_MOD, fd, &event);
-        switch (posix.errno(mod_rc)) {
-            .SUCCESS => {
-                entry.events = new_events;
-            },
-            else => |err| return unexpectedError(err),
-        }
-    }
+    // Keep the fd registered (edge-triggered, persistent). We intentionally do
+    // NOT epoll_ctl(DEL/MOD) here; the registration is torn down only when the
+    // socket is closed (see the net_close handler in submit). This eliminates
+    // the per-operation arm/disarm epoll_ctl churn that dominated this backend.
 }
 
 fn getHandle(completion: *Completion) NetHandle {
@@ -341,6 +293,12 @@ pub fn submit(self: *Self, state: *LoopState, c: *Completion) void {
             state.markCompletedFromBackend(c);
         },
         .net_close => {
+            const data = c.cast(NetClose);
+            // Tear down the persistent edge-triggered registration before the fd
+            // is closed, so a future fd that reuses this number starts clean.
+            if (self.poll_queue.fetchRemove(data.handle)) |_| {
+                _ = std.os.linux.epoll_ctl(self.epoll_fd, std.os.linux.EPOLL.CTL_DEL, data.handle, null);
+            }
             common.handleNetClose(c);
             state.markCompletedFromBackend(c);
         },
@@ -368,34 +326,99 @@ pub fn submit(self: *Self, state: *LoopState, c: *Completion) void {
             }
         },
 
-        // Other async operations - queue and try on wakeup
+        // Other async operations - try the syscall first (optimistic, like
+        // net_connect), only register with epoll on WouldBlock. This avoids the
+        // arm+disarm epoll_ctl pair when the socket is already ready.
         .net_accept => {
             const data = c.cast(NetAccept);
-            self.addToPollQueue(state, data.handle, c);
+            if (net.accept(data.handle, data.addr, data.addr_len, data.flags)) |handle| {
+                c.setResult(.net_accept, handle);
+                state.markCompletedFromBackend(c);
+            } else |err| switch (err) {
+                error.WouldBlock => self.addToPollQueue(state, data.handle, c),
+                else => {
+                    c.setError(err);
+                    state.markCompletedFromBackend(c);
+                },
+            }
         },
         .net_recv => {
             const data = c.cast(NetRecv);
-            self.addToPollQueue(state, data.handle, c);
+            if (net.recv(data.handle, data.buffers.iovecs, data.flags)) |n| {
+                c.setResult(.net_recv, n);
+                state.markCompletedFromBackend(c);
+            } else |err| switch (err) {
+                error.WouldBlock => self.addToPollQueue(state, data.handle, c),
+                else => {
+                    c.setError(err);
+                    state.markCompletedFromBackend(c);
+                },
+            }
         },
         .net_send => {
             const data = c.cast(NetSend);
-            self.addToPollQueue(state, data.handle, c);
+            if (net.send(data.handle, data.buffer.iovecs, data.flags)) |n| {
+                c.setResult(.net_send, n);
+                state.markCompletedFromBackend(c);
+            } else |err| switch (err) {
+                error.WouldBlock => self.addToPollQueue(state, data.handle, c),
+                else => {
+                    c.setError(err);
+                    state.markCompletedFromBackend(c);
+                },
+            }
         },
         .net_recvfrom => {
             const data = c.cast(NetRecvFrom);
-            self.addToPollQueue(state, data.handle, c);
+            if (net.recvfrom(data.handle, data.buffer.iovecs, data.flags, data.addr, data.addr_len)) |n| {
+                c.setResult(.net_recvfrom, n);
+                state.markCompletedFromBackend(c);
+            } else |err| switch (err) {
+                error.WouldBlock => self.addToPollQueue(state, data.handle, c),
+                else => {
+                    c.setError(err);
+                    state.markCompletedFromBackend(c);
+                },
+            }
         },
         .net_sendto => {
             const data = c.cast(NetSendTo);
-            self.addToPollQueue(state, data.handle, c);
+            if (net.sendto(data.handle, data.buffer.iovecs, data.flags, data.addr, data.addr_len)) |n| {
+                c.setResult(.net_sendto, n);
+                state.markCompletedFromBackend(c);
+            } else |err| switch (err) {
+                error.WouldBlock => self.addToPollQueue(state, data.handle, c),
+                else => {
+                    c.setError(err);
+                    state.markCompletedFromBackend(c);
+                },
+            }
         },
         .net_recvmsg => {
             const data = c.cast(NetRecvMsg);
-            self.addToPollQueue(state, data.handle, c);
+            if (net.recvmsg(data.handle, data.data.iovecs, data.flags, data.addr, data.addr_len, data.control)) |result| {
+                c.setResult(.net_recvmsg, result);
+                state.markCompletedFromBackend(c);
+            } else |err| switch (err) {
+                error.WouldBlock => self.addToPollQueue(state, data.handle, c),
+                else => {
+                    c.setError(err);
+                    state.markCompletedFromBackend(c);
+                },
+            }
         },
         .net_sendmsg => {
             const data = c.cast(NetSendMsg);
-            self.addToPollQueue(state, data.handle, c);
+            if (net.sendmsg(data.handle, data.data.iovecs, data.flags, data.addr, data.addr_len, data.control)) |n| {
+                c.setResult(.net_sendmsg, n);
+                state.markCompletedFromBackend(c);
+            } else |err| switch (err) {
+                error.WouldBlock => self.addToPollQueue(state, data.handle, c),
+                else => {
+                    c.setError(err);
+                    state.markCompletedFromBackend(c);
+                },
+            }
         },
         .net_poll => {
             const data = c.cast(NetPoll);
```
</details>

hyird:
Closes #510. Implements the idea in #262 (keep the fd in epoll), extended to be correct across event loops.

## Problem

The epoll backend is a one-shot, level-triggered reactor: `submit()` arms interest with `epoll_ctl(ADD)` and the fd is disarmed again with `epoll_ctl(DEL)` once it fires. One echo round-trip (recv + send) is two arm/disarm pairs, ~4 `epoll_ctl` per request — about 55% of kernel time in a `strace -f -c`. That's why the backend trailed `io_uring` by 25-35% on a plain TCP echo benchmark.

## Why "just keep the fd registered" isn't enough by itself

@lalinsky already pointed this out on #510: keeping the fd registered is only correct for a single event loop. zio runs one loop per executor and migrates fibers between them, so the *same* socket fd can be driven from different loops over its life:

- **migration mid-connection** — a fiber does `recv` on loop A, then migrates and does `send` on loop B;
- **concurrent full-duplex** — a reader task and a writer task on one socket land on different executors, so `recv` and `send` are in flight on different loops at once.

A naive per-loop "register once and never DEL" gets this wrong: when the fd is closed on one loop it leaves a stale `poll_queue` entry on the *other* loop, and when that fd number is reused there the backend thinks it's still registered, skips the ADD, and the new socket never gets events — a hang.

## Design

Interest is per-loop (each loop has its own epoll fd), so the loops in a group share a small table that records, per fd, **which loops currently have it registered**. The hot path stays off any per-op syscall, and there's no stale state to leave behind:

- **Optimistic syscall.** `submit()` tries `recv`/`send`/`accept` (and the `*from`/`*msg` variants) directly first, like the existing `net_connect` handler, and only registers on `WouldBlock`. This also guarantees the socket is drained to `EAGAIN` before we ever wait — which is what makes edge-triggered safe.
- **Persistent edge-triggered registration.** A socket is added once with `EPOLLIN|EPOLLOUT|EPOLLET` and stays registered for its lifetime (no `MOD` on read/write switches, no `DEL` on completion).
- **Shared registration table.** New `SharedState` = 64 mutex-sharded `fd -> u64 loop-bitmask` maps, with an IOCP-style `acquire`/`release` refcount that hands each loop a small `loop_bit`. `ensureRegistered(fd)` does the `epoll_ctl(ADD)` at most once per `(loop, fd)`, gated by the shared mask, so the same fd can be registered on two loops at once (cross-loop full-duplex) and each is tracked independently.
- **Ephemeral local entries.** A loop's `poll_queue` entry is dropped as soon as its completion list empties, so an idle socket never holds a local entry. Combined with `net_close` clearing the whole shared entry, this means a closed/reused fd number can never be mistaken for "already registered" — there's no stale state to reconcile across loops.
- **`net_poll` probe.** `net_poll` has no I/O to drain, so pure edge-triggered would hang on an already-(and continuously)-writable socket. It does a 0-timeout `poll()` readiness probe in `submit()` first and only waits on an edge when not ready.

The change is scoped to real sockets (the `net_*` ops). Other pollable fds — `pidfd` (process_wait), pipes, streaming files — are closed through paths that don't run `net_close`, so they stay on the original one-shot level-triggered path and never enter the shared table.

## Correctness

New in-tree test `multi-executor: cross-loop socket stress (full-duplex + migration + fd reuse)` (in `net.zig`): a 3-executor runtime drives many short connections in waves, each with its reader and writer as separate tasks (so recv/send for one fd run on different loops), and each echo handler spawns+joins a tiny task every round to force migration. Bytes are verified end-to-end; `std.testing.allocator` fails it on any leak, which also exercises the shared table's teardown. It passes on both `epoll` and `io_uring`.

Existing suite under `-Dbackend=epoll`: `net`, `Stream`, `recv`, `send`, `accept`, `connect`, `cancel`, `timeout`, `poll`, `process`, `pipe` all pass. DNS is 35/37 — the two failures are pre-existing live-DNS assertions and are identical on `main`.

## Benchmark

Closed-loop TCP ping-pong, 64-byte messages, system-default sockets (no `TCP_NODELAY`), server pinned with `taskset` so it's the bottleneck, 10s/run, 8-core EPYC 7402. req/s:

Single server thread:

| conns | epoll before | epoll after | io_uring |
|------:|-------------:|------------:|---------:|
| 50    | 93.8k        | **118.4k**  | 125.7k |
| 200   | 81.7k        | **99.9k**   | 102.2k |
| 1000  | 74.5k        | **86.0k**   | 88.3k |

3 server threads (3 executors):

| conns | epoll before | epoll after | io_uring |
|------:|-------------:|------------:|---------:|
| 50    | 266.5k       | **374.2k**  | 375.6k |
| 200   | 276.1k       | **377.4k**  | 400.3k |
| 1000  | 263.3k       | **365.8k**  | 374.7k |

`strace -f` on the single-thread server: `epoll_ctl` drops from ~4.0/req to **~0.0004/req** (52 ADDs total over 131k requests — one per socket, zero DEL/MOD). The per-op work on the wait path is now a sharded-map lookup instead of a syscall; the 3-executor numbers confirm it doesn't hurt multi-core scaling (epoll stays within a few percent of io_uring, essentially tied at 50 conns).

## Notes

- `kqueue` has the same per-op churn and the same single-loop limitation; this PR doesn't touch it, but the same shared-table approach applies and I'm happy to follow up.
- I arm `IN|OUT|ET` unconditionally, so there's one spurious `EPOLLOUT` wakeup per socket at registration (negligible).
- Teardown of the persistent registration happens in `net_close`; the kernel also drops a closed fd from epoll automatically, the explicit cleanup just keeps the tables tidy for fd reuse.

### DIFF
--- a/src/ev/backends/epoll.zig
+++ b/src/ev/backends/epoll.zig
@@ -65,47 +65,44 @@ const persist_events: u32 = linux.EPOLL.IN | linux.EPOLL.OUT | linux.EPOLL.ET;
 const reg_shard_count = 64; // power of two
 
 const RegShard = struct {
-    mutex: os.Mutex = .{},
+    mutex: os.Mutex = .init(),
     // fd -> bitmask of loop ids that have this socket registered in their epoll.
     map: std.AutoHashMapUnmanaged(NetHandle, u64) = .empty,
 };
 
 pub const SharedState = struct {
-    mutex: os.Mutex = .{},
-    refcount: usize = 0,
-    next_loop_id: u32 = 0,
+    mutex: os.Mutex = .init(),
+    /// Bit i set => loop id i is currently live. Ids are handed out from the
+    /// free (zero) bits and returned on release, so a group that churns loops
+    /// recycles ids instead of climbing past the 64-id budget.
+    used_mask: u64 = 0,
     allocator: std.mem.Allocator = undefined,
     shards: [reg_shard_count]RegShard = [_]RegShard{.{}} ** reg_shard_count,
 
-    /// Register a loop with the group. Returns a small stable id (0..63) used as
-    /// the loop's bit in the per-fd registration masks. Ids are recycled once the
-    /// group empties out (refcount back to zero), matching the executor lifetime.
+    /// Register a loop with the group. Returns a small id (0..63) used as the
+    /// loop's bit in the per-fd registration masks; the id is recycled once the
+    /// loop releases it.
     pub fn acquire(self: *SharedState, allocator: std.mem.Allocator) u6 {
         self.mutex.lock();
         defer self.mutex.unlock();
-        if (self.refcount == 0) {
-            self.allocator = allocator;
-            self.next_loop_id = 0;
-        }
-        const id = self.next_loop_id;
-        std.debug.assert(id < 64); // at most 64 executors share a group
-        self.next_loop_id += 1;
-        self.refcount += 1;
-        return @intCast(id);
+        if (self.used_mask == 0) self.allocator = allocator; // first loop in
+        const free = ~self.used_mask;
+        std.debug.assert(free != 0); // at most 64 executors share a group
+        const id: u6 = @intCast(@ctz(free));
+        self.used_mask |= @as(u64, 1) << id;
+        return id;
     }
 
-    pub fn release(self: *SharedState) void {
+    pub fn release(self: *SharedState, loop_bit: u64) void {
         self.mutex.lock();
         defer self.mutex.unlock();
-        std.debug.assert(self.refcount > 0);
-        self.refcount -= 1;
-        if (self.refcount == 0) {
+        self.used_mask &= ~loop_bit;
+        if (self.used_mask == 0) {
             // Last loop is gone; nobody can touch the shards now. Free them.
             for (&self.shards) |*shard| {
                 shard.map.deinit(self.allocator);
                 shard.map = .empty;
             }
-            self.next_loop_id = 0;
         }
     }
 
@@ -162,7 +159,7 @@ epoll_pwait2_supported: bool = true,
 
 pub fn init(self: *Self, allocator: std.mem.Allocator, queue_size: u16, shared_state: *SharedState) !void {
     const loop_id = shared_state.acquire(allocator);
-    errdefer shared_state.release();
+    errdefer shared_state.release(@as(u64, 1) << loop_id);
 
     const rc = std.os.linux.epoll_create1(std.os.linux.EPOLL.CLOEXEC);
     const epoll_fd: i32 = switch (posix.errno(rc)) {
@@ -210,7 +207,7 @@ pub fn deinit(self: *Self) void {
     if (self.epoll_fd != -1) {
         _ = std.os.linux.close(self.epoll_fd);
     }
-    self.shared.release();
+    self.shared.release(self.loop_bit);
 }
 
 pub fn wake(self: *Self, state: *LoopState) void {
--- a/src/ev/backends/epoll.zig
+++ b/src/ev/backends/epoll.zig
@@ -549,6 +549,24 @@ fn getHandle(completion: *Completion) NetHandle {
     };
 }
 
+/// Finish a completion that the optimistic syscall already satisfied, but via
+/// the loop's deferred completion queue instead of running its callback inline.
+///
+/// `submit` must never re-enter the submitter. The generic NetSendFile fallback
+/// (loop.zig) re-submits a NetSend from inside its own send callback, expecting
+/// the child to complete asynchronously like every other backend's net ops do.
+/// Completing inline there would re-enter `netSendFileAdvance` and finish the
+/// parent transfer twice. The syscall has already happened; only the completion
+/// notification is parked here and drained from `processCompletions`, off the
+/// submit call stack. `decrInflight` balances the `incrInflight` the loop did
+/// before calling `submit` (processCompletions runs `markCompleted`, which does
+/// not touch the inflight counter).
+fn completeDeferred(self: *Self, state: *LoopState, c: *Completion) void {
+    _ = self;
+    state.decrInflight();
+    state.work_completions.push(c);
+}
+
 /// Submit a completion to the backend - infallible.
 /// On error, completes the operation immediately with error.Unexpected.
 pub fn submit(self: *Self, state: *LoopState, c: *Completion) void {
@@ -638,12 +656,15 @@ pub fn submit(self: *Self, state: *LoopState, c: *Completion) void {
             const data = c.cast(NetSend);
             if (net.send(data.handle, data.buffer.iovecs, data.flags)) |n| {
                 c.setResult(.net_send, n);
-                state.markCompletedFromBackend(c);
+                // Defer, never complete inline: the NetSendFile fallback
+                // re-submits a send from inside its own send callback, and an
+                // inline completion would re-enter and finish the transfer twice.
+                self.completeDeferred(state, c);
             } else |err| switch (err) {
                 error.WouldBlock => self.addToPollQueue(state, data.handle, c),
                 else => {
                     c.setError(err);
-                    state.markCompletedFromBackend(c);
+                    self.completeDeferred(state, c);
                 },
             }
         },
--- a/src/ev/backends/epoll.zig
+++ b/src/ev/backends/epoll.zig
@@ -34,71 +34,104 @@ const BackendCapabilities = @import("../completion.zig").BackendCapabilities;
 
 pub const capabilities: BackendCapabilities = .{
     .process_wait = true,
+    // A socket fd is registered in exactly one loop's epoll (per direction), but
+    // the fiber that owns the operation can be on a different executor (task
+    // migration / cross-loop full-duplex). The registering loop completes the
+    // operation when its epoll fires, even though the completion was submitted on
+    // another loop, so the group's active/inflight accounting must be shared.
+    .is_multi_threaded = true,
 };
 
-// Persistent edge-triggered interest used for sockets. Both IN and OUT are armed
-// once for the socket's lifetime, so the backend never issues epoll_ctl(MOD) when
-// a socket switches between reading and writing, and never disarms on completion.
-// Edge-triggered is safe because submit() always tries the syscall first (draining
-// to EAGAIN) before it ever waits in epoll, so no readiness edge can be missed.
-const persist_events: u32 = linux.EPOLL.IN | linux.EPOLL.OUT | linux.EPOLL.ET;
-
-// --- Shared socket registration table -------------------------------------
+// --- Shared socket poll registry -------------------------------------------
 //
 // epoll interest is per-loop: an fd registered in loop A's epoll only produces
 // events on A's thread. zio runs one loop per executor and migrates fibers
-// between executors, so the *same* socket fd can be used from different loops
-// over its lifetime (migration), and a reader and a writer on one socket can
-// live on different loops at once (concurrent full-duplex). To keep the fd
-// registered (no per-op epoll_ctl) without leaving stale state behind, the loops
-// in a group share one table that records, per fd, which loops currently have it
-// registered in their epoll. A loop adds the fd to its own epoll the first time
-// it sees an operation for it and records its bit; on close the whole entry is
-// dropped (the kernel removes a closed fd from every epoll instance). The table
-// is sharded by fd to keep the common path off a single global lock.
+// between them, so a socket fd can be driven from different loops over its life,
+// and a reader and a writer can sit on different loops at once.
 //
-// Only true sockets (the net_* operations) go through this table. Other pollable
-// fds (pidfd for process_wait, pipes, streaming files) are closed through paths
-// that do not run net_close, so they stay on the original one-shot level-triggered
-// path where each operation arms and disarms its own interest.
+// The loops in a group share one table, keyed by fd, recording per direction
+// which loop owns the epoll registration plus the queue of completions waiting
+// on that (fd, direction). Lazy, single registration:
+//
+//   * submit does the optimistic syscall on whatever loop it lands on; on success
+//     it never touches the table (lock-free fast path).
+//   * on WouldBlock it looks up (fd, dir): if a loop already owns it, the
+//     completion is just parked as a waiter (no epoll_ctl); otherwise the current
+//     loop becomes the owner and registers fd in its own epoll.
+//   * when the owner's epoll fires it drains the waiters - completing each, even
+//     those submitted on other loops, via the shared accounting above.
+//
+// This registers the fd on the few loops that actually poll it (one per
+// direction) rather than on every loop a migrating fd visits, and delegating the
+// completion to the loop that already has it registered avoids an extra
+// registration syscall on the parking loop.
+//
+// Only true sockets (the net_* operations) use this table. Other pollable fds
+// (pidfd for process_wait, pipes, streaming files) are closed through paths that
+// do not run net_close, so they stay on the original one-shot level-triggered
+// per-loop path.
 
 const reg_shard_count = 64; // power of two
 
+const Dir = enum(u1) { in = 0, out = 1 };
+
+/// Per-fd registration state, shared across the loops in a group.
+const FdReg = struct {
+    /// Loop whose epoll has fd registered for read / write (null = not yet).
+    owner_in: ?*Self = null,
+    owner_out: ?*Self = null,
+    /// Completions parked waiting for readability / writability.
+    waiters_in: Queue(Completion) = .{},
+    waiters_out: Queue(Completion) = .{},
+    /// Readiness latch. Set when an edge fires with no waiter parked yet (a
+    /// "wasted" edge), so the next op to park sees it and re-checks instead of
+    /// waiting for an edge that edge-triggered epoll will not repeat. This closes
+    /// the race where data arrives between an op's optimistic syscall returning
+    /// EAGAIN and that op enqueuing as a waiter.
+    ready_in: bool = false,
+    ready_out: bool = false,
+
+    fn empty(self: *const FdReg) bool {
+        return self.owner_in == null and self.owner_out == null and
+            self.waiters_in.empty() and self.waiters_out.empty();
+    }
+};
+
 const RegShard = struct {
     mutex: os.Mutex = .init(),
-    // fd -> bitmask of loop ids that have this socket registered in their epoll.
-    map: std.AutoHashMapUnmanaged(NetHandle, u64) = .empty,
+    map: std.AutoHashMapUnmanaged(NetHandle, FdReg) = .empty,
 };
 
 pub const SharedState = struct {
     mutex: os.Mutex = .init(),
-    /// Bit i set => loop id i is currently live. Ids are handed out from the
-    /// free (zero) bits and returned on release, so a group that churns loops
-    /// recycles ids instead of climbing past the 64-id budget.
-    used_mask: u64 = 0,
+    refcount: usize = 0,
     allocator: std.mem.Allocator = undefined,
+
+    /// Multi-threaded counters: a completion submitted on one loop may be
+    /// finished by the loop that owns the fd registration, so active/inflight_io
+    /// are shared atomics rather than per-LoopState fields.
+    active: std.atomic.Value(u64) = .init(0),
+    inflight_io: std.atomic.Value(u64) = .init(0),
+
     shards: [reg_shard_count]RegShard = [_]RegShard{.{}} ** reg_shard_count,
 
-    /// Register a loop with the group. Returns a small id (0..63) used as the
-    /// loop's bit in the per-fd registration masks; the id is recycled once the
-    /// loop releases it.
-    pub fn acquire(self: *SharedState, allocator: std.mem.Allocator) u6 {
+    pub fn acquire(self: *SharedState, allocator: std.mem.Allocator) void {
         self.mutex.lock();
         defer self.mutex.unlock();
-        if (self.used_mask == 0) self.allocator = allocator; // first loop in
-        const free = ~self.used_mask;
-        std.debug.assert(free != 0); // at most 64 executors share a group
-        const id: u6 = @intCast(@ctz(free));
-        self.used_mask |= @as(u64, 1) << id;
-        return id;
+        if (self.refcount == 0) {
+            self.allocator = allocator;
+            self.active.store(0, .release);
+            self.inflight_io.store(0, .release);
+        }
+        self.refcount += 1;
     }
 
-    pub fn release(self: *SharedState, loop_bit: u64) void {
+    pub fn release(self: *SharedState) void {
         self.mutex.lock();
         defer self.mutex.unlock();
-        self.used_mask &= ~loop_bit;
-        if (self.used_mask == 0) {
-            // Last loop is gone; nobody can touch the shards now. Free them.
+        std.debug.assert(self.refcount > 0);
+        self.refcount -= 1;
+        if (self.refcount == 0) {
             for (&self.shards) |*shard| {
                 shard.map.deinit(self.allocator);
                 shard.map = .empty;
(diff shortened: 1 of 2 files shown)


lalinsky:
@hyird looks good conceptually, I'll need a bit more time to review, but can you take a look at the failing test? it seems that operations are marked as completed without setting the result in some cases

lalinsky:
Ok, I missed one key part, you are registering the poll on whetever epoll the recv/send lands on, with task migration, single fd could potentially be registered on ALL event loops. I'm not sure if that's good.

My original idea was this:

- We have a shared poll queue, keyed by `(fd, direction)`, so at most two entries per socket.
- We register with epoll lazily, on the first recv/send.
- Handling send/recv:
   * Do the optimistic send/recv on whatever event loop you are on, if it worked, you resume task and don;t even touch the poll queue.
   * If you get `WouldBlock`, then you look up if the fd/dir is already registered.
       + If yes, you add your completion to the wait queue, and return (park yourself). This needs careful synchronization.
      + If no, you add to the shared poll queue and register with your local epoll.

So this approach is supported:

- recv previously ran on executor A
- now it gets migrated to executor B and it runs recv again
- if recv syscall succeeds, it gets resumed on executor B and continues normally
- if recv syscall got `EAGAIN`, you register as waiter, and executor A will resume the task locally when the event fires

I think this is strictly better, because you are not registering the fd on all event loops, just the key ones. We already use task migration by default, so recv being delivered on executor A when it started on executor B is not an issue.

Actually, when the user disabled task migration, we should probably not even use the shared loop group, so this solution would also work with that case and fds would only ever be register locally.

Now, I'm not sure whether to accept your version, or push for this approach, but I think it is better.

Key points:
- fast-path touches no lock, needs no migration
- if we need to poll, it's strictly always faster to just delegate the completion to the loop that already has it registered, because it will arrive sooner, than if we registered is ourselved on the local epoll (less syscalls, lower latency)
- if we keep the direction as part of the key, we support reader/writer on different threads without any issue
- lazy registration means the balancing depends on the scheduler, and we are not forcing all sockets to just one epoll

What do you think?

hyird:
Agreed, your approach is better. Registering on whichever loop the op lands on means a migrating fd ends up in every epoll — wasteful, and it causes spurious wakeups on loops that no longer handle the socket. Keying the shared queue by `(fd, direction)` and registering once is cleaner, and reader/writer on different threads falls out for free. I'll rework the PR to it.

One thing I'd like to pin down before building it: when the fd is registered on loop A but the parked completion came from loop B, how should it finish —

- A completes it directly (cross-loop), which means active/inflight accounting has to be shared across the group, like the IOCP backend's counters; or
- A just wakes B and B finishes it locally (per-loop accounting, one extra hop).

I'm leaning toward the first, since that's the "A resumes the task locally when the event fires" path you described — but it means epoll opts into the shared-counter path. And as you said, I'll skip the shared loop group entirely when task migration is disabled, so single-threaded stays fully local with no cross-loop machinery at all. Does that match what you had in mind?

(Aside: the failing test is fixed on the current branch — the optimistic send was completing inline and re-entering the NetSendFile fallback, finishing the transfer twice; I deferred that completion, CI is green now. That code changes in the rework anyway.)

lalinsky:
Correct, when the event is delivered on A, we handle it on A, even though the request came from B. That is very similar to the IOCP path, indeed. Everything should already be able to handle it, but it's worth double-checking it. The epoll backend should be then declared as `multi_threaded` in capabilities, which just changes some internal accounting, everything else is already thread-safe.

hyird:
Pushed the rework: shared `(fd, direction)` poll queue, lazy single registration, and cross-loop completion via the multi_threaded path. The owning loop drains and completes waiters that were submitted on other loops, exactly like IOCP, and epoll is now `multi_threaded` in capabilities.

One thing worth flagging from double-checking it: there's a wakeup race between an op's optimistic syscall returning EAGAIN and it parking as a waiter. If data arrives in that window and the owner's edge fires before the waiter is enqueued, edge-triggered epoll won't repeat the edge. I close it with a per-`(fd, dir)` readiness latch: the owner latches an edge that found no waiter, so the next op to park re-checks the syscall instead of waiting.

Full suite is green (same 482/484 as main, the 2 are the pre-existing live-DNS tests), including the multi-executor cross-loop stress test (full-duplex reader/writer on different executors + migration + fd reuse, byte-verified).

Heads up on the cost: every blocking op now takes the sharded queue lock, and the shared counters are atomic even single-threaded, so on a non-migrating echo benchmark this is a few percent behind the previous "register locally" version (still within ~5-10% of io_uring single-thread; multi-thread is on par). The win is on the migration / cross-loop side, which that benchmark doesn't exercise.

hyird:
Addressed both review comments in 38f89d8:

- **Close orphaning parked ops** (critical): `unregisterSocket` now drains `waiters_in`/`waiters_out` and completes each with `error.Canceled` (deferred, off the close's submit stack) before dropping the entry, so a close racing a pending op on another loop no longer hangs the fiber or unbalances the accounting. The normal close path (no parked waiters) is unchanged.
- **parkSocket inline error completions** (minor): the registry-OOM and `epoll_ctl`-failure paths now go through `completeDeferred` like the success path, so a `net_send` reaching them from the NetSendFile fallback can't complete inline and re-enter.

Full suite still 482/484 (the 2 are the pre-existing live-DNS tests), cross-loop stress green.

src/ev/backends/epoll.zig
```diff
@@ -163,26 +277,290 @@ fn getEvents(completion: *Completion) u32 {
...
+                    shard.mutex.unlock();
+                } else {
+                    shard.mutex.unlock();
+                    c.setError(error.Canceled);
+                    self.completeDeferred(state, c);
+                }
+            },
+        }
+        return;
+    }
```
lalinsky, line 466: **Lost wakeup in `parkSocket`'s latch-recheck → requeue path**

The recheck path consumes `ready`, drops the shard lock to run the syscall, and on `EAGAIN` re-parks as a waiter **without re-reading `ready`** under the re-acquired lock. During that unlocked syscall window the owner loop's `drainDir` can latch `ready = true` for data that just arrived — and because the registration is edge-triggered, that already-buffered data produces no further edge, so the re-parked completion sleeps until the *next* arrival.

Interleaving (P parking on loop B, A = owner of `(fd, in)`):

```
1. P: ready == true -> clear ready; unlock
2. P: recv() -> EAGAIN              (socket momentarily empty)
3. A: data arrives, epoll edge fires
4. A: drainDir: no waiter queued yet (P not re-enqueued),
      !drained_any -> ready = true
5. P: re-lock; push P as waiter; unlock   // never re-checks ready
   -> P parked, ready == true, data buffered, no future edge -> hang
```

The precondition (a *stale* latch) is reachable on its own: a latched `ready` consumed by an optimistic `submit()` syscall never clears the flag, so the next op finds `ready == true` and enters the recheck path. With a per-op timeout this surfaces as a spurious `error.Timeout`; an untimed op hangs. The tell is the asymmetry — `drainDir`'s own requeue clears `ready` under the lock (epoll.zig:515), but this branch doesn't even read it.

Root cause: the syscall must run outside the lock, so the recheck path drops and re-acquires — and the decision to park is torn across that gap. The invariant that needs restoring is *the enqueue must be guarded by a `ready == false` observed in the same critical section.*

**Fix** — make the recheck a loop; only re-run the syscall when `ready` is set, so the common `ready == false` park costs no extra syscall:

```zig
var ev = std.os.linux.epoll_event{
    .data = .{ .fd = fd },
    .events = switch (dir) {
        .in => std.os.linux.EPOLL.IN,
        .out => std.os.linux.EPOLL.OUT,
    },
};
while (true) {
    shard.mutex.lock();
    const r = shard.map.getPtr(fd) orelse {        // closed underneath us
        shard.mutex.unlock();
        c.setError(error.Canceled);
        self.completeDeferred(state, c);
        return;
    };
    // net_connect can't distinguish success/failure without the real event,
    // so it never rechecks — it just parks (matches the current carve-out).
    if (c.op == .net_connect or !readyPtr(r, dir).*) {
        waitersPtr(r, dir).push(c);                // ready false -> park, no syscall
        shard.mutex.unlock();
        return;
    }
    readyPtr(r, dir).* = false;
    shard.mutex.unlock();
    c.prev = null;
    c.next = null;
    switch (checkCompletion(c, &ev)) {
        .completed => {
            self.completeDeferred(state, c);
            return;
        },
        .requeue => {},                            // re-loop: re-test ready under the lock
    }
}
```

Now the enqueue can no longer slip past a concurrently-set latch — after every recheck syscall we re-take the lock and re-test `ready` before being allowed to park. Caveats: re-fetch `reg`/`ready`/`waiters` after each re-lock (a concurrent `getOrPut` on another fd can rehash the shard map). Termination is bounded — a recheck only repeats if `ready` was re-latched, i.e. a real edge fired, which the syscall then consumes.

The first-registration path (owner becomes self) doesn't need this: a fresh `EPOLLET` `ADD` reports current readiness as an edge, and `ready` can't be set while `owner` is `null`.

(Note: this is narrow — it needs cross-loop ownership and a specific interleaving, so the stress test stayed green across many runs. But it's a genuine lost wakeup, not just a wasted syscall.)

lalinsky:
Thanks a lot for this, @hyird — your shared cross-loop registration approach is what drove the design here.

I've taken it forward in #526, which reworks the same idea into a single-owner per-`(fd, direction)` registration shared across both epoll and kqueue. Closing this in favor of that PR, but your branch is merged into #526's history (with authorship preserved + a `Co-authored-by` trailer), so the contribution is credited when #526 lands.

Really appreciate the work and the review iterations here.

PR #512, opened 2026-06-23
random: async-native zio.random / zio.randomSecure

lalinsky:
Implements native `zio.random` and `zio.randomSecure`, aligned with the `std.Io` interface, and drops the previous delegation to `std.Io.Threaded` for entropy.

## Design

Two primitives (`src/random.zig`):

- **`random`** — fast, non-cryptographic bytes from a **per-`Executor` CSPRNG** (`std.Random.DefaultCsprng`), seeded once at executor startup and then filling with **zero syscalls**. One executor per thread → lock-free (a `random()` call never suspends mid-fill). The CSPRNG is purely an async-path optimization; outside an executor it falls back to the secure syscall.
- **`randomSecure`** — cryptographically secure entropy obtained fresh from the OS on every call. The blocking syscall is delegated to a thread-pool worker (via `blockInPlace`) so the event loop is never stalled; runs inline when called outside a task.

We **own the syscalls** in `src/os/` (`os.getrandom`):

| Platform | Mechanism |
|---|---|
| Linux | `getrandom(2)` |
| BSD / macOS | `arc4random_buf` (kernel-seeded; what std uses with libc) |
| Windows | `ProcessPrng` (the modern API used by Go/Rust/Chromium) |

The `ev` event-loop layer is intentionally **not** involved — the CSPRNG lives on `Executor`, not on `Loop`.

## Left for later (per-backend async, skip the thread pool)

- io_uring: async read of `/dev/urandom` (char device → io-wq punt, no zio pool slot)
- Windows: `NtDeviceIoControlFile(\Device\CNG, IOCTL_KSEC_GEN_RANDOM)` via IOCP

## Notes

- The Windows `ProcessPrng` path is compile-dispatched but unexercised on the Linux dev box (the `randomSecure` test skips Windows).
- Full suite: 485/487, with only the two pre-existing unrelated `setSize` failures.

## Commits
1. per-loop/executor non-secure CSPRNG (`zio.random`)
2. own secure syscalls (`zio.randomSecure`)
3. seed CSPRNG from own RNG, drop std dependency
4. move CSPRNG to `Executor` and API to `src/random.zig`


src/os/posix.zig
```diff
@@ -316,3 +316,36 @@ pub const EFD = @import("eventfd.zig").EFD;
...
+                    else => return error.EntropyUnavailable,
+                }
+            }
+        },
+        else => {
+            // BSD / macOS: arc4random_buf is the platform-recommended CSPRNG
+            // (kernel-seeded via getentropy) and cannot fail. This is the same
+            // source std.Io.Threaded uses when libc is linked.
+            if (@hasDecl(system, "arc4random_buf")) {
+                system.arc4random_buf(buffer.ptr, buffer.len);
```
lalinsky, line 345: `arc4random_buf` always fills the full buffer of fails? no errors?

src/random.zig
```diff
@@ -0,0 +1,139 @@
...
+const runtime = @import("runtime.zig");
+const common = @import("common.zig");
+const Cancelable = common.Cancelable;
+const blockInPlace = common.blockInPlace;
+const getCurrentExecutorOrNull = runtime.getCurrentExecutorOrNull;
+
+/// Per-executor non-cryptographic CSPRNG. Owned by `Executor`, seeded once at
+/// startup from a secure source. Fills are synchronous and perform no syscalls.
+/// The seed must come from a cryptographically secure source.
+pub const Csprng = struct {
```
lalinsky: What is the purpose of this empty wrapper?

src/runtime.zig
```diff
@@ -329,6 +334,15 @@ pub const Executor = struct {
         try setupStackGrowth();
         errdefer cleanupStackGrowth();
 
+        // Seed this executor's non-secure CSPRNG with an independent secure
+        // seed drawn from zio's own blocking secure RNG. Each executor draws
+        // its own seed so per-thread streams are independent. If the OS cannot
+        // provide entropy at startup, executor creation fails rather than
+        // running with a predictable seed.
+        var seed: [random_mod.Csprng.seed_len]u8 = undefined;
```
lalinsky: This should be a `pub fn` in random.zig, but not exposed via `zio`, e.g. `random_mod.setup()`

src/random.zig
```diff
@@ -0,0 +1,139 @@
...
+/// state. The blocking syscall is delegated to a thread-pool worker so the
+/// event loop is not blocked; when called outside a task it runs inline.
+///
+/// Future per-backend optimizations can skip the thread pool entirely: async
+/// reads from /dev/urandom on io_uring, and \Device\CNG via IOCP on Windows.
+pub fn randomSecure(buffer: []u8) RandomSecureError!void {
+    // Result lives on our stack, defaulted to Canceled, so that if the work is
+    // canceled before it runs we surface cancellation instead of a bogus value.
+    var result: RandomSecureError!void = error.Canceled;
+    blockInPlace(struct {
```
lalinsky, line 60: You can't return value from `blockInPlace`, right?

src/os/posix.zig
lalinsky: Correct — `arc4random_buf` returns void and cannot fail; it always fills the entire buffer (OpenBSD/macOS guarantee it never fails, no fd, no error path). That is also why `std.Io.Threaded` calls it directly with no error check.

src/random.zig
lalinsky: Agreed, removed in 1222ec8. Now using `std.Random.DefaultCsprng` directly (exposed as the `random.Csprng` alias). The forwarding struct added nothing: we eagerly seed at executor startup, so there is no uninitialized sentinel to model.

src/runtime.zig
lalinsky: Done in 1222ec8 — added `pub fn setup()` in random.zig (not re-exported via `zio`). `Executor.init` now just calls `random_mod.setup()`.

src/random.zig
lalinsky: `blockInPlace` does return a value, but on pre-run cancellation (work canceled before the worker starts) its result is left `undefined`, so returning it directly would be unsafe. I capture into a stack slot defaulted to `error.Canceled` instead, which also gives correct cancellation semantics. Happy to switch to a dedicated thread-pool wrapper if you prefer that over the out-param.

src/random.zig
```diff
@@ -0,0 +1,139 @@
...
+/// back to the secure syscall directly (best effort: zeros the buffer if
+/// entropy is unavailable).
+///
+/// For cryptographically secure entropy, use `randomSecure` instead.
+pub fn random(buffer: []u8) void {
+    if (getCurrentExecutorOrNull()) |exec| {
+        exec.csprng.fill(buffer);
+    } else {
+        os.getrandom(buffer) catch @memset(buffer, 0);
+    }
```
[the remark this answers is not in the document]
lalinsky: Already addressed in 1222ec8 (this review raced the push). The no-executor path no longer zeros on failure: it seeds an ad-hoc `std.Random.DefaultPrng` from the monotonic clock (always available), so callers never get predictable bytes.

src/random.zig
```diff
@@ -0,0 +1,140 @@
...
+}
+
+/// The per-executor non-cryptographic CSPRNG type.
+pub const Csprng = std.Random.DefaultCsprng;
+
+/// Create a freshly-seeded per-executor CSPRNG from OS entropy. Called by the
+/// runtime at executor startup; not part of the public `zio` API. Each executor
+/// gets an independent seed so per-thread streams are independent. Fails if the
+/// OS cannot provide entropy rather than running with a predictable seed.
+pub fn setup() os.GetRandomError!Csprng {
```
lalinsky: Actually, I change my mind, let's have `RandomState` in this file, and it's `setup(&state)`, so that we can optimize with the opened `/dev/random` later on.

lalinsky: Done in c4901f2 — `RandomState` struct in random.zig, initialized via `setup(&state)`. It only holds the CSPRNG for now; the `/dev/urandom` fd and Windows CNG handle can be added there later so the secure path can skip the thread pool.

PR #515, opened 2026-06-23
Wall-clock timers: per-clock heaps + std.Io sleep/futex/batch integration

lalinsky:
Stacked on #513 (`clock-resolution`).

Makes `awake`/`boot`/`real` timers behave distinctly and wires the clock through every `std.Io` timeout, so boot/real sleeps and absolute realtime deadlines work end-to-end. Per-backend native arming (io_uring REALTIME/BOOTTIME, etc.) is intentionally **not** here — that's the next PR.

## Per-clock timer heaps (`ev`)
- `ev.Timer` carries a `clock`; timers live in one heap per wall-clock domain (`awake`/`boot`/`real`), since their deadlines are in different epochs and only comparable within a clock. CPU-time clocks are rejected (panic).
- `Clock` gets explicit contiguous values so it indexes the heaps and the cached `now[]` directly via `@intFromEnum`.
- `now` is cached per clock: `awake` refreshed eagerly each scan, `boot`/`real` lazily on first use and stamped with a scan counter — an idle heap reads no clock, an active one reads at most once per scan.
- No backend arms wall clocks natively yet, so `boot`/`real` fall back to folding their remaining into the poll timeout, bounded by `wall_clock_cap` (10s) so they re-evaluate across suspend/clock-steps. The per-poll re-read of `now(clock)` is what makes the cap bound oversleep.

## std.Io integration
The vtable has exactly three timeout entry points; all now thread the clock:
- **`sleep`** — `sleepImpl` carries the clock; `fromStd` now keeps a deadline as an absolute deadline (in its clock's epoch) instead of flattening it to a monotonic-relative duration. The old conversion subtracted the monotonic now from a possibly-foreign epoch, which mishandled realtime deadlines.
- **`futexWait`** — underlies every timed sync primitive (`Mutex.timedLock`, `Condition.timedWait`, `Event.timedWait`, timed `Future.await`); wiring it covers all of them.
- **`batchAwaitConcurrent`** — both the multi-op wait loop and the single-op fast path (`operateInner`) thread the clock.

Threading uses `*Clock` variants (`Waiter.timedWaitClock`, `Futex.timedWaitClock`, `timedWaitForIoClock`) that the existing `.awake` callers delegate to unchanged, plus `Clock.fromStd` / `Clock.fromStdTimeout` helpers.

## Testing
- ev: `boot` (duration) and `real` (absolute deadline) timers fire via the fallback path.
- std.Io: `io.sleep` on `boot`/`real`, and sleeping until an absolute realtime deadline (would have hung at `Duration.max` before the `fromStd` fix).
- Full suite green (492/492); Windows (wine) and macOS compile verified earlier in the series.

## Known minor gaps (acceptable, noted in comments)
- The no-task `timedWaitFutex` fallback is monotonic-only, so boot/real degrade to awake when there's no fiber.
- The frozen-slice `now` is anchored at sleep-time, not wake-time (deliberate).

## Next PR
Per-backend native wall-clock arming (io_uring `IORING_TIMEOUT_ABS|REALTIME/BOOTTIME`, epoll timerfd, IOCP waitable timer, kqueue `EVFILT_TIMER`), gated by a `native_wall_timers` capability that makes `checkTimers` stop folding boot/real into the poll timeout for that backend.

### DIFF
--- a/src/common.zig
+++ b/src/common.zig
@@ -8,6 +8,7 @@ pub const log = std.log.scoped(.zio);
 
 const ev = @import("ev/root.zig");
 const Timeout = @import("time.zig").Timeout;
+const Clock = @import("time.zig").Clock;
 const Stopwatch = @import("time.zig").Stopwatch;
 const Runtime = @import("runtime.zig").Runtime;
 const getCurrentTaskOrNull = @import("runtime.zig").getCurrentTaskOrNull;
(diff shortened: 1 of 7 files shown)


lalinsky:
@coderabbitai review

PR #523, opened 2026-06-25
task: add TaskLocal task-local storage

lalinsky:
Adds `zio.TaskLocal(T)` — per-task key/value storage, the task-scoped analogue of a thread-local. Closes the "store data on the current task" gap without `context.Context`-style explicit threading.

## API

Manual binding with caller-provided node storage:

```zig
var trace_id: zio.TaskLocal(u64) = .{};

fn handler() void {
    var node: @TypeOf(trace_id).Node = .unset;
    trace_id.set(&node, 42);
    defer trace_id.clear(&node);
    // trace_id.get() returns 42 here and in any callee,
    // across yields and executor migration.
}
```

Or `scoped` for a single call (no node to manage):

```zig
const out = trace_id.scoped(42, handler, .{args});
```

- `set(node, value)` — bind a value using **caller-provided** node storage.
- `clear(node)` — remove a binding and return the node to `.unset` (reusable). May interleave with other bindings in any order, not just LIFO.
- `get()` — `?T`, the innermost active binding by value, or null if unbound / outside a task. For mutable per-task state, bind a pointer (`TaskLocal(*T)`) and mutate through it.
- `scoped(value, func, args)` — bind for the dynamic extent of one call, then restore. The node lives on the stack (no allocation) and `func` is invoked via `@call`, so it takes any comptime-known callable (`args: std.meta.ArgsTuple(@TypeOf(func))`), not a fn pointer.

## Design

- **Keyed by the instance address.** `&trace_id` is the key, so multiple instances of the same `T` are distinct, and a library can declare its own without any central registry. (The type carries a 1-byte marker so distinct instances get distinct addresses — a zero-sized type would let two instances share one.)
- **Caller-owned node storage ⇒ no allocation, no cleanup policy.** The binding lives exactly as long as the node: stack node for a scoped binding, embedded/heap node for a task-lifetime one. No per-task map, no allocator involvement, no registry.
- **Intrusive doubly-linked chain on `AnyTask`.** Carried across task migration for free (the chain hangs off the task, not the executor), and only ever touched by the owning task, so no locking. Bindings nest and shadow.
- **`get` returns by value.** Handing out `*T` into the node would dangle once the binding is cleared or its frame returns, and leak internal storage; mutable state is expressed by binding a pointer instead.
- **`scoped` uses a comptime callable + `@call`**, not a fn pointer, so the convenience form allocates nothing either.

## Node lifecycle & safety

Nodes are declared `.unset` (a known unbound state) rather than `undefined`. The node's `key` doubles as the "is it linked?" marker, which makes the misuse checks cheap and read no uninitialized memory:

- The node must stay alive and unmoved between `set` and `clear`.
- `set` asserts the node is `.unset` (O(1)) — catches reusing a still-bound node.
- `clear` asserts the node is bound by this `TaskLocal` — catches a stray clear-before-set and double-clear (both leave the node `.unset`), then resets it to `.unset` for reuse.
- A task that **returns with a binding still live** trips an assert in safety builds — catching a forgotten `clear` at the task that leaked it.

## Tests

Eight tests in `task.zig`: set/get/clear by value, mutable state via a bound pointer, nested shadowing/restore, non-LIFO middle-node unlink, `scoped` bind/restore + return value, node reuse after clear, per-task isolation across a yield, and unbound-key/distinct-instance lookup. Green across all CI targets including 32-bit (riscv32/arm/thumb). (Any lone red on `release riscv32` is a pre-existing qemu flake in `ev.ThreadPool: one task` — eventfd `.BADF` — unrelated to this change.)

## Open questions for review

- Naming: `set`/`clear`/`get`/`scoped`, `TaskLocal`/`Node`/`.unset`.

### DIFF
--- a/src/task.zig
+++ b/src/task.zig
@@ -654,7 +654,7 @@ pub fn TaskLocal(comptime T: type) type {
         /// The node lives on the stack, so this allocates nothing, and `func` is
         /// invoked via `@call` (any comptime-known callable, no fn-pointer
         /// indirection). Returns whatever `func` returns.
-        pub fn scoped(self: *Self, value: T, func: anytype, args: anytype) @TypeOf(@call(.auto, func, args)) {
+        pub fn scoped(self: *Self, value: T, func: anytype, args: std.meta.ArgsTuple(@TypeOf(func))) @TypeOf(@call(.auto, func, args)) {
             var node: Node = .unset;
             self.set(&node, value);
             defer self.clear(&node);
--- a/src/task.zig
+++ b/src/task.zig
@@ -167,8 +167,10 @@ pub const CancelKind = enum { user, auto };
 /// which executor currently runs the task (tasks keep their identity across
 /// migration).
 pub const TaskLocalNode = struct {
-    /// The `TaskLocal` instance this binding belongs to (its address is the key).
-    key: *const anyopaque = undefined,
+    /// The `TaskLocal` instance this binding belongs to (its address is the key),
+    /// or null when the node is unbound. Doubles as the "is this node linked?"
+    /// marker, so `set`/`clear` can catch misuse cheaply.
+    key: ?*const anyopaque = null,
     /// The task whose chain this node is linked into.
     owner: *AnyTask = undefined,
     prev: ?*TaskLocalNode = null,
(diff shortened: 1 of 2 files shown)


lalinsky:
@coderabbitai review

PR #527, opened 2026-06-26
DO NOT MERGE: repro cross-loop stress crash (netbsd/windows), trimmed CI

lalinsky:
Throwaway diagnostic. Adds **only** the multi-executor cross-loop stress test to `main` (no backend changes) with CI trimmed to NetBSD + Windows, to isolate whether the crash is pre-existing in the runtime's cross-thread/task machinery or introduced by the single-owner socket backend (#526). Will close once located.


lalinsky:
Served its purpose: confirmed the cross-loop stress crash is a race-group use-after-free in `Group.groupCallback` (decrement-before-cancel), exposed by concurrent cross-thread completion of a timed op's two children. Fixed in #528 — full CI green on NetBSD **and** Windows (IOCP hit the same race). Closing the diagnostic branch.

PR #528, opened 2026-06-26
ev/completion: fix race-group UAF when children complete on different loops

lalinsky:
Stacked on #526. Fixes the cross-loop race-group use-after-free that surfaced as the NetBSD-debug segfault in #526's cross-loop stress test (fault value == a freed timer's timeout).

`Group.groupCallback` decremented `remaining` before the race-winner walked `self.head` to cancel siblings. Once children complete concurrently on different loops (any `is_multi_threaded` backend — IOCP, and epoll/kqueue with cross-loop socket servicing), the thread that drives `remaining` to 0 completes the group and frees the stack frame while the winner is still iterating → UAF. Fix: cancel siblings **before** decrementing.

Root-caused by reading the code + the decisive #527 experiment (the test passes on `main` (not multi_threaded) and only fails with the cross-loop backend). Likely also fixes the Windows/IOCP failure (IOCP is already multi_threaded). Verifying on full CI, NetBSD especially.

### DIFF
--- a/src/ev/completion.zig
+++ b/src/ev/completion.zig
@@ -417,9 +417,21 @@ pub const Group = struct {
 
     fn groupCallback(loop: *Loop, completion: *Completion) void {
         const self: *Group = @ptrCast(@alignCast(completion.group.owner.?));
-        const prev = self.remaining.fetchSub(1, .acq_rel);
 
-        // Race mode: first to clear the flag wins, cancels siblings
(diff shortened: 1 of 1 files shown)


lalinsky:
Superseded by #529, which landed the same groupCallback fix (plus the stress test) directly on main. The backend PR #526 will pick it up on rebase.

Issue #530, opened 2026-06-26
IOCP: intermittent lost wakeup under cross-loop full-duplex + migration


lalinsky:
## Summary

Under a heavy multi-executor workload — concurrent full-duplex on a single socket (reader and writer tasks on different executors), task migration, and fd reuse — the **IOCP backend intermittently loses a socket completion**: an in-flight `recv`/`send`/`accept` never completes, so the op blocks until its timeout (or forever, if untimed).

This is **distinct from** the race-group use-after-free fixed in #529 (that was a crash, in shared `completion.zig`, and is fixed). This one is a **lost wakeup** specific to the **IOCP** backend, and it's flaky/timing-dependent.

## Symptom

The `multi-executor: cross-loop socket stress (full-duplex + migration + fd reuse)` test (added in #529, `src/net.zig`) intermittently fails on `windows-latest` / `iocp`:

- the test runs ~8 s instead of ~250 ms (an op sits until the 5 s `io_timeout`),
- then fails its `sh.errors == 0` assertion (the timeout is counted as an error).

So a completion was never delivered — i.e. a lost wakeup, not a crash.

## Frequency

Flaky. Observed across recent CI runs of the same test on Windows/IOCP:

| run | windows debug | windows release |
|---|---|---|
| #528 | pass | pass |
| #529 | pass | pass |
| #526 (rebased) | pass | **fail (8 s timeout)** |

So roughly 1-in-several on `release` (ReleaseSafe) so far; `debug` has passed throughout. (A rerun is in flight to refine the rate.)

## Scope / cause

- The IOCP backend is **unchanged** by the epoll/kqueue single-owner work (#526) — that code is comptime-gated to those backends. IOCP is already `is_multi_threaded` on `main`, so the test exercises its existing cross-thread completion path. This is therefore a **pre-existing latent IOCP bug**, newly surfaced by the stress test.
- The workload that triggers it: a connection's **reader and writer run as separate tasks on different executors** (so recv and send for one fd are in flight on different IOCP loops at once), plus per-round `spawn`+`join` migration and fd reuse across waves.
- Likely area: IOCP cross-thread completion delivery / association when a single fd is driven from more than one loop concurrently. Worth checking the recv/send completion-port association and the wakeup path for the case where the op was submitted on one loop and its completion is dequeued on another.

## Impact

Because the stress test now lives on `main` (#529), **`main`'s Windows CI can intermittently fail** until this is fixed. Options in the meantime:
[...]

PR #532, opened 2026-06-26
DEBUG: reproduce + instrument IOCP cross-loop stall (#530) — do not merge

lalinsky:
Throwaway debug branch to reproduce and diagnose #530 in CI. Enables the cross-loop stress test on IOCP, cranks load, lowers io_timeout to 3s, and logs STALL (with CancelIoEx outcome) + LATE completions. Matrix trimmed to Windows/IOCP. Not for merge.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -46,7 +46,7 @@ jobs:
           -Demit-test-bin
         BIN=$(ls zig-out/bin/test* | head -1)
         echo "Looping $BIN"
-        for i in $(seq 1 40); do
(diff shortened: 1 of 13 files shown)


lalinsky:
Closing — root cause found and fixed in #533 (a submit/complete data race: the accept socket was stored into result_private *after* AcceptEx, so on the shared multi-threaded IOCP port a cross-thread completion could read an uninitialized handle and double-close it). All the instrumentation here served its purpose; the fix landed minimally on main.

PR #535, opened 2026-06-30
blocking_task: fix reference count leak in registerBlockingTask

jrc2139:
## Problem

`spawnBlocking` leaks the blocking-task allocation. `registerBlockingTask` increments `awaitable.ref_count`, but the equivalent `registerTask` (regular tasks) does not. That extra reference is never released:

- the init ref (=1 from `RefCounter.init`) is dropped by `finishTask` via `threadPoolCompletion`;
- the caller / `JoinHandle` ref (added in `spawnBlockingTask` before submit) is dropped by `join()` / `cancel()` / `detach()`;
- the extra `incr()` in `registerBlockingTask` has no matching release, so `AnyBlockingTask.destroy()` is never called.

## Why it wasn't caught

The leak is masked for pool-sized tasks: `Runtime.deinit` → `TaskPool.deinit` frees the pool's backing buffer regardless of whether individual tasks were destroyed. It only becomes observable for tasks whose allocation exceeds `pool_item_size` and take the direct `rt.allocator.alignedAlloc` path. The existing `spawnBlocking` smoke test returns an `i32`, so its task is pool-allocated and the leak stays hidden.

## Fix

Drop the extra `ref_count.incr()` in `registerBlockingTask` so it mirrors `registerTask`.

## Test

Adds `runtime: spawnBlocking does not leak with a large result` (returns `[4000]u8` to force the direct-allocator path). On `main` it reports `1 test leaked`; with the fix it passes. Full suite: 502/502 tests pass.

### DIFF
--- a/src/blocking_task.zig
+++ b/src/blocking_task.zig
@@ -129,7 +129,13 @@ fn registerBlockingTask(rt: *Runtime, task: *AnyBlockingTask) error{RuntimeShutd
         return error.RuntimeShutdown;
     }
 
-    task.awaitable.ref_count.incr();
+    // Do NOT increment ref_count here. Unlike registerTask (task.zig), this would
+    // add a third reference that is never released: the init ref (=1) is dropped by
+    // finishTask via threadPoolCompletion, and the caller (JoinHandle) ref is added
+    // in spawnBlockingTask before submit and dropped by join()/cancel()/detach().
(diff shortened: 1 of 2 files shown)


lalinsky:
@claude analyze this change, explain what are the refs for in each task type

lalinsky:
Thank you

PR #538, opened 2026-07-03
feat: add support for io_uring ATTACH_WQ

### LINKED ISSUE #486: Use ATTACH_WQ for io_uring setup
We can share some resources, will probably need event loops to register in the loop group, so that we have access to their ring fd.

notxorand:
closes: #486

### DIFF
--- a/src/ev/backends/io_uring.zig
+++ b/src/ev/backends/io_uring.zig
@@ -200,7 +200,6 @@ wall_generation: [2]u32 = .{ 0, 0 },
 /// SQE until the next `io_uring_enter2` reads it.
 wall_ts: [2]linux.kernel_timespec = undefined,
 shared_state: *SharedState,
-is_master: bool = false,
 
 pub fn init(self: *Self, allocator: std.mem.Allocator, queue_size: u16, shared_state: *SharedState) !void {
     var flags: u32 = 0;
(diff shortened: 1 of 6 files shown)


lalinsky:
The master fd should be in SharedState, which is a backend-specific struct, not directly in LoopGroup. This is similar to how we have just one shared IOCP handle, for example. And we should also have refcount there. So that the master fd won't get closed until the last event loop needs it.

notxorand:
yes. what I decided to do was set the master ring's fd to -1 (inside the io_uring.deinit) in order to keep the fd open till we no longer have any ring using it. does that work?

src/ev/backends/io_uring.zig
```diff
@@ -196,25 +199,48 @@ wall_generation: [2]u32 = .{ 0, 0 },
...
-    _ = shared_state;
     var flags: u32 = 0;
     flags |= linux.IORING_SETUP_SINGLE_ISSUER;
     flags |= linux.IORING_SETUP_DEFER_TASKRUN;
     flags |= linux.IORING_SETUP_COOP_TASKRUN;
 
-    var ring = try linux.IoUring.init(queue_size, flags);
+    var ring = blk: {
+        if (shared_state.master_fd.load(.seq_cst) != -1) {
+            const master_fd = shared_state.master_fd.load(.seq_cst);
```
lalinsky: Why do you load it twice? You should have have this load at the beginning of the block and then `if (master_fd != -1) { ... }`

src/ev/backends/io_uring.zig
```diff
@@ -223,9 +249,28 @@ pub fn init(self: *Self, allocator: std.mem.Allocator, queue_size: u16, shared_s
...
+        .sq_thread_idle = 1000,
+        .wq_fd = @as(u32, @intCast(master_fd)),
+    });
+
+    return try linux.IoUring.init_params(queue_size, &params);
+}
+
 pub fn deinit(self: *Self) void {
     _ = linux.close(self.waker_eventfd);
+    if (self.is_master) self.ring.fd = -1;
```
lalinsky: Technically, `is_master` is redundant. 

You could have done this:

```
const master_fd = self.shared_state.master_fd.load(.seq_cst);
if (self.ring.fd == master_fd) self.ring.fd = -1;
```

Is there another reason to have the field?

notxorand: no other reason. removing that

notxorand: also, the -1 trick seemed to fail. so, I decided to manually offload the master ring's resources

lalinsky (approved):

lalinsky:
Thank you!

PR #541, opened 2026-07-04
feat: work stealing

### LINKED ISSUE #460: Implement work stealing for load balancing across executors
## Background

The scheduler currently distributes load in two ways:

- **Spawn-time placement:** new tasks are assigned a home executor round-robin (`getNextExecutor`).
- **Migration:** an already-running task that gets woken on a different executor migrates to that executor (the "waker runs the wakee" heuristic, when `enable_task_migration` is set).

There is no rebalancing once tasks are placed. This means load can become uneven for some workloads — e.g. producer/consumer or fan-out patterns can concentrate work on a single executor while others sit idle, with nothing to pull the backlog elsewhere.

## Proposal

Add work stealing so idle executors can pull ready tasks from busy ones, giving the scheduler a way to rebalance after the initial placement.

## Cleanup once this lands

`scheduleTask` currently keeps new tasks (`old.tag == .new`) pinned to their round-robin assigned home executor and explicitly excludes them from migration, specifically to preserve distribution in the absence of rebalancing. Once work stealing exists, that `old.tag != .new` check can be removed and new tasks can follow the same locality-first scheduling as everything else, letting the stealer handle distribution.

notxorand:
closes: #460


src/utils/concurrent_stack.zig
```diff
@@ -38,6 +38,25 @@ pub fn ConcurrentStack(comptime T: type) type {
             }
         }
 
+        /// Atomically pops an item from the stack.
+        /// Returns null if the stack is empty.
+        pub fn pop(self: *Self) ?*T {
+            while (true) {
+                const current_head = self.head.load(.acquire);
+                if (current_head == null) return null;
+                const new_head = current_head.?.next;
```
lalinsky: This operation is not safe. Another thread could pop, run, finish, destroy the task between line 44 and 46. You could be reading `next` from undefined memory at this point. See my comments on the original issue regarding pop safety.

src/runtime.zig
```diff
@@ -739,6 +787,8 @@ pub const Runtime = struct {
     options: RuntimeOptions,
 
     executors: std.ArrayList(*Executor) = .empty,
+    global_queue: SimpleQueue(WaitNode) = .empty,
+    global_queue_mutex: Mutex = .init,
```
lalinsky: You can't use zio async `Mutex` here, you need `os.Mutex` here.

src/runtime.zig
```diff
@@ -427,17 +427,38 @@ pub const Executor = struct {
                 @panic("event loop stopped while the main task was yielding");
             }
 
-            // Drain remote ready queue (cross-thread tasks) after processing current queue
-            var drained = self.next_ready_queue_remote.popAll();
-            while (drained.pop()) |task| {
-                self.ready_queue.push(task);
-                self.ready_count += 1;
+            for (&self.queue_pool) |*stack| {
```
lalinsky: Why are you reading all the queues, just to redistribute them? Plus, you are pushing to queues you will drain next. This makes little sense to me, please explain.

src/runtime.zig
```diff
@@ -427,17 +427,38 @@ pub const Executor = struct {
...
+            var has_local_work = false;
+            for (&self.queue_pool) |*stack| {
+                if (stack.head.load(.acquire) != null) {
+                    has_local_work = true;
+                    break;
+                }
+            }
+
+            if (!has_local_work) {
+                if (self.stealWork()) continue;
```
lalinsky: You shouldn't continue without running I/O poll at least once.

src/runtime.zig
```diff
@@ -451,59 +472,98 @@ pub const Executor = struct {
...
+                self.runtime.global_queue_mutex.unlock();
+            }
         }
 
-        // Peek at head of ready_queue
-        const node = self.ready_queue.head orelse return null;
-        const task = AnyTask.fromWaitNode(node);
+        // Try to pop from the current cursor and rotate if empty
+        for (0..num_stealable_stacks) |_| {
+            if (self.queue_pool[self.pop_cursor].pop()) |node| {
```
lalinsky: Key part of my design was that we read one stack at a time, if we empty it, we do I/O poll, then rotate. That's the main justification for the LIFO stacks being OK. If you read from all the stacks, especially if you are comparing against `last_run_tick`, these queue MUST be FIFO, which makes the whole atomic dance harder, to the point where the Go-style atomic ring buffer + overflow queue is a much better design.

src/runtime.zig
```diff
@@ -451,59 +472,98 @@ pub const Executor = struct {
...
+        // Maximum tasks to run before forcing an event loop tick
         const EVENT_INTERVAL = 61;
-
-        // Force event loop tick after running EVENT_INTERVAL tasks
-        if (self.tick_task_count >= EVENT_INTERVAL) {
-            return null;
+        if (self.tick_task_count >= EVENT_INTERVAL) return null;
+
+        // Try to pop from the global queue occasionally
+        if (self.tick_task_count % 8 == 0) {
```
lalinsky: Popping from global queue should happen when local queue (one of the four) is empty. Or at the 61 interval. Watch this video why the 61 was chosen. It's a prime, so it doesn't have any patterns. That's why 8 is not a good number for this purpose.

https://www.youtube.com/watch?v=-K11rY57K7k

src/runtime.zig
```diff
@@ -427,17 +427,38 @@ pub const Executor = struct {
...
 
             // Run event loop - non-blocking if there's work, otherwise wait for I/O
             const main_ready = check_ready and self.main_task.state.load(.acquire).tag == .ready;
-            const has_work = self.ready_queue.head != null or main_ready;
-            try self.loop.run(if (has_work) .no_wait else .once);
+            if (!has_local_work) has_local_work = main_ready;
+
+            if (!has_local_work) {
+                self.runtime.global_queue_mutex.lock() catch unreachable;
+                has_local_work = self.runtime.global_queue.head != null;
```
lalinsky: While this is nice, mergeable PR will need to wake up sleeping event loop once new work arrives.

lalinsky (commented):
This is a critical core part of the library, so be prepared that I'll be hard on the review. :)

Nothing, but a perfect solution, will be accepted.

src/runtime.zig
notxorand: it was a relic from my old approach where the ready_queue was still a part of the design

src/runtime.zig
```diff
@@ -739,6 +778,8 @@ pub const Runtime = struct {
     options: RuntimeOptions,
 
     executors: std.ArrayList(*Executor) = .empty,
+    global_queue: SimpleQueue(WaitNode) = .empty,
+    global_queue_mutex: std.Io.Mutex = .init,
```
lalinsky: You misunderstood, this is still async mutex, if you use `rt.io()`. I meant `os.Mutex` from `src/os/thread.zig`.

src/runtime.zig
```diff
@@ -451,59 +466,95 @@ pub const Executor = struct {
...
         // Maximum tasks to run before forcing an event loop tick (from Go's scheduler)
         const EVENT_INTERVAL = 61;
-
-        // Force event loop tick after running EVENT_INTERVAL tasks
-        if (self.tick_task_count >= EVENT_INTERVAL) {
-            return null;
+        if (self.tick_task_count >= EVENT_INTERVAL) return null;
+
+        // Try to pop from the global queue occasionally
+        if (self.tick_task_count % EVENT_INTERVAL == 0) {
```
lalinsky: You don't need to do this check here, simply handle it in Executor.run after you are done with one batch of tasks.

src/utils/concurrent_stack.zig
notxorand: is the two-swap pop satisfactory here?

notxorand:
@lalinsky, I think the mutex might need a way to let us know it didn't succeed. e.g., a situation where we previously locked and we're attempting to lock again. really, the only issue might be logic bugs, but it seems to happen (ran into one today 😅)

src/runtime.zig
```diff
@@ -988,6 +1045,21 @@ pub const Runtime = struct {
...
+        // Try to claim the "searcher" token so only one thread wakes up
+        if (self.searchers.cmpxchgStrong(0, 1, .acq_rel, .acquire) == null) {
+            // We armed a searcher. Now, pick an idle executor to wake.
+            // We use trailing zeros to pick the first bit set in the mask.
+            const target_id = @ctz(idle);
+            if (target_id < self.executors.items.len) {
+                self.executors.items[target_id].loop.wake();
+            }
+        }
+    }
```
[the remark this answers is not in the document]
notxorand: don't we risk clearing `searchers` if we weren't the searcher?

notxorand:
will keep at it in the morning. the failing ci, some seem to be unrelated but others seem to be an issue with using a 64 bit atomic with 32 bit archs? (I'm unsure). how do we deal with failing ci? and tell me what part of this needs improvements

src/runtime.zig
```diff
@@ -247,36 +249,34 @@ pub fn getNextExecutor(rt: *Runtime) error{RuntimeShutdown}!*Executor {
...
     last_tick_time: Timestamp = .zero,
 
     // Deferred cleanup for the task that just yielded away from this executor.
-    // Processed by the next coroutine to run (at landing sites: startFn, yield resume, run loop).
     pending_cleanup: TaskCleanup = .none,
 
-    // Remote task support - lock-free LIFO stack for cross-thread resumption
-    next_ready_queue_remote: ConcurrentStack(WaitNode) = .{},
+    // Local task support - lock-free LIFO stacks for cross-thread resumption and stealing.
+    queue_pool: [num_stealable_stacks]ConcurrentStack(WaitNode) = .{ .{}, .{}, .{}, .{} },
```
lalinsky: Use `@splat` here.

src/runtime.zig
```diff
@@ -402,7 +402,7 @@ pub const Executor = struct {
     const yield_ready_threshold = 13;
 
     pub fn maybeYield(self: *Executor, comptime mode: AnyTask.YieldMode, comptime cancel_mode: YieldCancelMode) if (cancel_mode == .allow_cancel) Cancelable!void else void {
-        if (self.ready_count >= yield_ready_threshold) {
+        if (self.ready_count.load(.seq_cst) >= yield_ready_threshold) {
```
lalinsky: `seq_cst` is unwaranted here, why?

src/runtime.zig
```diff
@@ -451,17 +451,37 @@ pub const Executor = struct {
                 @panic("event loop stopped while the main task was yielding");
             }
 
-            // Drain remote ready queue (cross-thread tasks) after processing current queue
-            var drained = self.next_ready_queue_remote.popAll();
+            var drained = self.queue_pool[self.pop_cursor].popAll();
```
lalinsky: Why do you do this drain?

src/runtime.zig
```diff
@@ -475,60 +495,126 @@ pub const Executor = struct {
...
+        self.push_cursor = (self.push_cursor + 1) % num_stealable_stacks;
+        _ = self.ready_count.fetchAdd(1, .acq_rel);
     }
 
     /// Schedule a task to a remote executor (different executor or no current executor).
     /// Uses the thread-safe remote queue and notifies the executor.
-    fn scheduleTaskRemote(self: *Executor, task: *AnyTask) void {
+    pub fn scheduleTaskRemote(self: *Executor, task: *AnyTask) void {
         std.debug.assert(task != &self.main_task);
+        std.debug.assert(self.runtime.options.enable_task_migration);
```
lalinsky: Keep this for now, we will need to find a way for this to work without task migration, maybe we will add local inject queue to drain instead of the global queue in that mode.

src/runtime.zig
```diff
@@ -563,34 +649,23 @@ pub const Executor = struct {
             break;
         }
 
+        if (getCurrentExecutorOrNull()) |current_exec| {
```
lalinsky: By reordering this, you made it possible to schedule another executor's main_task as a regular task to our executor.

src/runtime.zig
```diff
@@ -475,60 +495,126 @@ pub const Executor = struct {
...
-        // Task already ran this tick? Force event loop tick first.
-        // This prevents a yielding task from running multiple times per tick.
-        // We leave the task in the queue (don't pop) to preserve FIFO order.
-        if (task.last_run_tick == self.current_tick) {
-            return null;
+        if (self.tick_task_count >= EVENT_INTERVAL) return null;
+
+        // Try to pop from the current cursor stack and rotate if empty
+        // We check all stacks to ensure we don't block on I/O while work is pending.
+        for (0..num_stealable_stacks) |_| {
```
lalinsky: I've already said this, but you can't do this. You can only read tasks from one queue at a time, OR you need to switch them to FIFO.

Two problems:

1) last_run_tick check with LIFO stacks make no sense, they are not in the right order

2) if you remove last_run_tick check, then you have no fairness policy

src/runtime.zig
```diff
@@ -1518,27 +1616,3 @@ test "runtime: mutex contention with task migration" {
     try std.testing.expect(!group.hasFailed());
     try std.testing.expectEqual(2_000, counter);
 }
-
-test "runtime: disable main executor" {
```
lalinsky, line 1647: I guess you had to delete this due to the bug I pointed out in scheduleTask.

src/runtime.zig
```diff
@@ -475,60 +495,126 @@ pub const Executor = struct {
...
+                for (&victim.queue_pool) |*stack| {
+                    var drained = stack.popAll();
+                    var count: u32 = 0;
+                    while (drained.pop()) |task| {
+                        self.queue_pool[self.push_cursor].push(task);
+                        self.push_cursor = (self.push_cursor + 1) % num_stealable_stacks;
+                        count += 1;
+                    }
+
+                    _ = self.ready_count.fetchAdd(count, .acq_rel);
```
lalinsky: You never subtract the victim's ready_count after stealing.

src/runtime.zig
```diff
@@ -475,60 +495,126 @@ pub const Executor = struct {
...
+                has_work = self.popWorkFromGlobal();
+            }
+        }
+        return has_work;
+    }
+
+    fn popWorkFromGlobal(self: *Executor) bool {
+        if (self.runtime.options.enable_task_migration) {
+            self.runtime.global_queue_mutex.lock();
+            if (self.runtime.global_queue.pop()) |node| {
```
lalinsky: When taking from global, you should always take more tasks, size/num_executors would be a good chunk.

src/runtime.zig
```diff
@@ -788,6 +863,11 @@ pub const Runtime = struct {
     options: RuntimeOptions,
 
     executors: std.ArrayList(*Executor) = .empty,
+    global_queue: SimpleQueue(WaitNode) = .empty,
+    global_queue_mutex: OsMutex = .init(),
+    idle_mask: u64 = 0,
+    idle_mask_mutex: OsMutex = .init(),
```
lalinsky: You could reduce max_executors to 32 on 32-bit platforms and then use u32, but that can way, the mutex is fine for now.

src/runtime.zig
```diff
@@ -451,17 +451,37 @@ pub const Executor = struct {
...
-                self.ready_count += 1;
+                self.queue_pool[self.push_cursor].push(task);
+                self.push_cursor = (self.push_cursor + 1) % num_stealable_stacks;
+            }
+
+            // Set ourselves as idle before scanning for work.
+            const my_bit = @as(u64, 1) << @as(u6, @intCast(self.id));
+            if (self.runtime.options.enable_task_migration) {
+                self.runtime.idle_mask_mutex.lock();
+                self.runtime.idle_mask |= my_bit;
```
lalinsky: This will need more work, you can't set the bit all the time. If you still have tasks to run, it's a lie to mark yourself as idle. It can lead to all kinds of issues.

lalinsky (changes requested):

src/utils/concurrent_stack.zig
lalinsky: Yes. The swap+store is an improvement over my original idea. It's fine, but I don't think it should live in ConcurrentStack, because it depends on single producer, and the producer being the same as the pop-er. So maybe a specific struct for the run queue would be better.

src/runtime.zig
notxorand: it's slow and doesn't require as much consistency?

lalinsky:
I think we will need to change course a little bit. I'm sorry to say this, but I don't have much confidence that you know what you are doing too well here. Like, the last commit, introducing a non-concurrent doubly-linked list and removing the last_tick fairness policy, is very wrong.

Can you please write down the design of what you want to end up with? And then follow that design.

src/runtime.zig
notxorand: just for clarification, what I meant here was that I understand that .seq_cst should be replaced with something less expensive

src/runtime.zig
notxorand: left-over logic. a lot of the code was me originally starting out with replacing `next_ready_queue_remote` with the `queue_pool` and just forgetting to adjust to match

lalinsky:
I've done a LOT of benchmarking, all possible variants of concurrent linked lists, plus the Go-style ring buffer, and I must say, the ring buffer is by far the cheapest option. It's super close to the current plain-ptr linked list. The atomic linked lists are 1.5-2x slower in a single channel ping-pong benchmark.

This is why I wanted to do it in phases, because it allows us to decide on the queue before reworking the whole scheduler. It's not possible to get relevant numbers if you mix too many changes.

If you want, I can open a PR for the ring buffer that essentially implements phase 1 of my original design. It just replaces the current queues, but doesn't add work stealing yet. And then you can work on the design of the idle/searching executors + work stealing.

notxorand:
> If you want, I can open a PR for the ring buffer that essentially implements phase 1 of my original design. It just replaces the current queues, but doesn't add work stealing yet. And then you can work on the design of the idle/searching executors + work stealing.

yes, do

lalinsky:
I've merged the queue swap, plus had to fix a nasty bug where LLVM was silently ignoring our register clobbers and the queue swap exposed it.

notxorand:
work stealing requires a new pr and branch or we'll just have to squash the merge

PR #542, opened 2026-07-05
Fix use-after-free in blockInPlace cancellation

lalinsky:
## Problem

The `common.test.blockInPlace: cancellation interrupts a blocking syscall on the worker` test segfaults on powerpc64le/epoll under QEMU in CI (reported crash: `cmpxchgStrong` at address `0x0`, under the `completion_fn` call in `thread_pool.runTasks`). It reproduces fairly regularly in CI.

## Root cause

`blockInPlace` keeps everything on its own stack — `ctx`, `waiter`, `token`, `work`. The invariant that keeps this safe is: the task must not return (and reclaim that stack) until the worker's `completionFn` has completely finished touching those objects.

The cancel handler broke that invariant. `completionFn` publishes `done` **before** its trailing `waiter.signal()`:

```zig
ctx.done.store(true, .release);   // (1)
waiter.signal();                  // (2) still reads our stack-allocated waiter
```

The cancel loop returned as soon as it observed `done`:

```zig
while (!ctx.done.load(.acquire)) { ... }
return ctx.result;   // frame reclaimed here
```

So the sequence was:

1. Worker sets `ctx.done = true`.
2. The canceled task (parked in the resend-backoff `timedWait`) is woken by the backoff timer, sees `done == true`, and returns from `blockInPlace` — **reclaiming its stack**.
3. Worker executes `waiter.signal()` against the now-freed `waiter`. The reused stack memory makes the tagged-union `mode` read as `.select`, which takes the `cmpxchgStrong` path (`common.zig`, `Waiter.signal`) on a garbage `winner` pointer → segfault at `0x0`.

The window between (1) and (2) is only a couple of instructions on the worker, so the crash requires the backoff timer to fire in exactly that gap. That is rare on fast native hardware but happens regularly under QEMU emulation and CI load. Nothing here is architecture-specific — it's a pure timing window.

The fast (non-canceled) path was never affected: it returns only after `signal()`, which is `completionFn`'s last action there.

## Fix

Keep resending `SIGURG` with backoff until the work finishes (`done`), but drive the backoff sleeps on a **separate** waiter, leaving `blockInPlace`'s `waiter` to be signaled only by `completionFn`. After the loop, do one final `waiter.wait(1, .no_cancel)` to synchronize with `completionFn`'s trailing `signal()` — the worker's last access to the stack frame — before returning.

`completionFn` and the fast path are unchanged.

### DIFF
--- a/src/common.zig
+++ b/src/common.zig
@@ -411,8 +411,11 @@ pub fn blockInPlace(func: anytype, args: std.meta.ArgsTuple(@TypeOf(func))) meta
     const Context = struct {
         args: Args,
         result: Result = undefined,
-        // Distinct from the waiter's signal count: the resend timer below also
-        // signals the waiter, so we need an unambiguous "work finished" flag.
+        // Set by completionFn (before it signals the waiter) to tell the cancel
+        // path's resend loop that the work has finished, so it stops resending.
+        // It is NOT a "safe to return" flag: completionFn still touches our stack
+        // frame in the trailing waiter.signal(), which the cancel path waits for
+        // separately. See the cancel handler below.
         done: std.atomic.Value(bool) = .init(false),
 
         fn workFn(work: *ev.Work) void {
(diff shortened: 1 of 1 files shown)


lalinsky:
not the right fix

lalinsky:
Superseded by #543, which fixes the same UAF more cleanly by waiting for the work through the event loop (loop-owned completion delivery) instead of the custom completion callback + manual resend loop. #543 is stacked on #539.

PR #546, opened 2026-07-05
Route panic and scheduler I/O through the blocking path (#545)

### LINKED ISSUE #545: Panic while inside Loop.add recurses through debug_io and aborts, masking the original panic
## Summary

When a panic occurs while the current task is inside `Loop.add` / an in-flight I/O op, and the program has set `std_options_debug_io = zio.debug_io`, the panic handler writes the panic message to stderr **through zio's async I/O**. That write goes `debug.defaultPanic → lockStderr → flush → Io.File.Writer.drain → io.fileWriteStreamingImpl → waitForIo → Loop.add`. But we're already inside `Loop.add` (its `in_add` guard is set, and the `defer` that clears it hasn't run because we're unwinding a panic, not returning). So `Loop.add` hits:

```zig
if (self.in_add) {
    @panic("recursive call to Loop.add() is not allowed");
}
```

That panic tries to report itself the same way → `Loop.add` → panic → ... infinite recursion → `abort()`. The process dies with SIGABRT and **no panic message**, so the original panic is completely hidden.

## Backtrace (abbreviated, recursion trimmed)

```
#4  abort
#5  process.abort
#6  debug.defaultPanic
#7  ev.loop.Loop.add                      <-- panics: "recursive call to Loop.add()"
#8  common.waitForIo
#9  common.timedWaitForIoClock
#10 io.fileWriteStreamingImpl
...
#15 Io.File.Writer.drainStreaming (io.stderr_writer)
#17 Io.Writer.defaultFlush
#19 Io.LockedStderr.clear
#20 Io.lockStderr
#21 debug.lockStderr
#22 debug.defaultPanic                     <-- reporting the *previous* panic
#23 ev.loop.Loop.add                       <-- ...which was also inside Loop.add
   ... repeats ...
```

## Impact

Any panic that happens to unwind while a task is inside `Loop.add` (or any op that reaches it) is turned into an unrelated `recursive call to Loop.add()` abort with no message. This makes real bugs very hard to diagnose — you can't see what actually panicked. Encountered while debugging a normal app that sets `std_options_debug_io`.

## Suggested fix

The panic path must not perform async, loop-driven I/O. When writing a panic message, `debug_io` (or the panic hook) should fall back to a **direct blocking write** to stderr (e.g. plain `write(2)`), never `waitForIo`/`Loop.add`. Alternatively, detect re-entry (`in_add` / in-panic) and route the panic report through the blocking path.

## Workaround

Don't set `std_options_debug_io = zio.debug_io` (panics then print via the default blocking writer). This also reveals the true original panic.

lalinsky:
Fixes #545.

## Problem

When a panic unwinds while a task is inside `Loop.add` (or any in-flight I/O op) and the app sets `std_options_debug_io = zio.debug_io`, writing the panic message goes back through zio's async I/O:

```
defaultPanic -> lockStderr -> drain -> fileWriteStreamingImpl -> waitForIo -> Loop.add
```

`Loop.add`'s `in_add` guard is still set (its clearing `defer` never ran — we're unwinding, not returning), so it panics `"recursive call to Loop.add()"`, which reports itself the same way → infinite recursion → `abort()` with **no panic message**. The original panic is completely masked.

The same trap applies to any `std.log` / `std.debug.print` emitted from scheduler code (e.g. a completion callback, or `yield()` internals).

## Fix

`waitForIo` already falls back to a synchronous, in-thread syscall (`executeBlocking`) whenever there is no current task. This makes that condition true exactly where the loop must not be re-entered:

- `current_task` becomes `?*AnyTask` and is cleared while the executor is running scheduler code — the run loop between steps, and `switchOut` while it picks the next task / switches. It is set back to a task before every switch into a task body, and restored to `main_task` when the run loop returns.
- The crash handler (`crashHandlerImpl`) clears `current_task` so panic-message I/O always takes the blocking path, even when unwinding from *inside* `Loop.add`. This is sticky — once crashing, we never touch the loop again.

Forcing the blocking path surfaced a latent gap: std's stack-trace symbolization opens, reads and then **closes** the executable through `debug_io`, and both `fileCloseImpl` and `netCloseImpl` batch closes into a gather `Group`, which `executeBlocking` used to `@panic` on. `executeBlocking` now runs a gather group's children sequentially, after validating each child is an op it can run inline (`file_close` / `net_close`); other ops keep the clear "requires event loop" panic.

## Verification

- All unit tests pass (`./check.sh`), all examples build (`zig build examples`).
- Manual repro: a task that panics with `std_options_debug_io = zio.debug_io` set. Before this change the blocking path also broke on the grouped close (recursive panic during symbolization); with the full fix it prints the complete panic message **and** symbolized stack trace via the blocking path, then aborts normally — and because all panic-time I/O now bypasses the loop, the same holds when the panic unwinds from inside `Loop.add`.

### DIFF
--- a/src/runtime.zig
+++ b/src/runtime.zig
@@ -284,10 +284,14 @@ pub const Executor = struct {
     // Periodic timer for evicting idle stacks from the shared stack pool.
     stack_pool_eviction_timer: ev.Timer = ev.Timer.init(.{ .duration = .fromSeconds(60) }),
 
-    // The task currently executing on this executor.
+    // The task currently executing on this executor, or null when we are running
+    // scheduler code (the run loop, a context switch, a completion callback) rather
+    // than a task body. A null value routes any I/O performed here through the
+    // blocking path in waitForIo instead of re-entering the event loop, which would
+    // otherwise recurse into Loop.add (see issue #545).
     // Updated before every context switch into a task and after every switch back.
     // Used by getCurrentTaskOrNull() instead of the TLS current_context chain.
-    current_task: *AnyTask,
+    current_task: ?*AnyTask,
 
     // Executor dedicated to this thread. Written once on init, never updated.
     pub threadlocal var current_DO_NOT_ACCESS_DIRECTLY: ?*Executor = null;
@@ -406,6 +410,16 @@ pub const Executor = struct {
     pub fn run(self: *Executor, mode: RunMode) !void {
         const check_ready = mode != .until_stopped;
 
+        // The run loop is scheduler code, not a task: clear current_task so any I/O
+        // it performs (a completion callback, std.log, a debug_io write) takes the
+        // blocking path rather than recursing into the event loop. Restored to the
+        // main task on return, since control goes back to the main task's user code
+        // (or, for worker executors, to shutdown). This defer does not run on the
+        // crash path: markCrashed()'s callers are noreturn and a panic does not
+        // unwind through defers, so the null marker it sets survives until abort().
+        self.current_task = null;
+        defer self.current_task = &self.main_task;
+
         // Process deferred cleanup (e.g. main task's park/reschedule)
         self.processCleanup();
 
@@ -415,7 +429,7 @@ pub const Executor = struct {
                 @atomicStore(*Context, &next_task.coro.parent_context_ptr, &self.main_task.coro.context, .release);
                 self.current_task = next_task;
                 next_task.coro.step();
-                self.current_task = &self.main_task;
+                self.current_task = null;
                 self.processCleanup();
             }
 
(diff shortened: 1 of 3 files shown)


src/runtime.zig
```diff
@@ -406,6 +410,16 @@ pub const Executor = struct {
     pub fn run(self: *Executor, mode: RunMode) !void {
         const check_ready = mode != .until_stopped;
 
+        // The run loop is scheduler code, not a task: clear current_task so any I/O
+        // it performs (a completion callback, std.log, a debug_io write) takes the
+        // blocking path rather than recursing into the event loop. Restored to the
+        // main task on return, since control goes back to the main task's user code
+        // (or, for worker executors, to shutdown). This defer does not run on the
+        // crash path: markCrashed()'s callers are noreturn and a panic does not
+        // unwind through defers, so the null marker it sets survives until abort().
```
lalinsky, line 419: The described failure mode can't occur in Zig. `markCrashed()` is only reached from `crashHandlerImpl` → std's `defaultPanic` / `defaultHandleSegfault`, both of which are `noreturn` and end in `abort()`. A panic (or segfault handler) does not unwind the stack running `defer` blocks, so control never returns to `run()` after `markCrashed()` — its exit `defer` never executes, and the null marker survives until `abort()`.

I verified this empirically: with a task that panics under `debug_io`, the symbolization close took the *blocking* path (it hit `executeBlocking`'s group case), which only happens if `current_task` stayed null the whole time — i.e. the `run()` defer did not restore it.

So a `crashed` bit would guard an unreachable path. Instead I clarified the comments on both `run()` and `markCrashed()` to state exactly why the restore can't run on the crash path (noreturn callers + no defer unwinding on panic).

PR #548, opened 2026-07-05
runtime: expose a top-level maybeYield()

lalinsky:
## What

Adds a top-level `zio.maybeYield()`.

## Why

`Executor.maybeYield` already exists — it yields only when enough tasks are waiting (`ready_count >= yield_ready_threshold`), so it's a cheap fairness check rather than an unconditional context switch. But it's a method on the internal `Executor` and wasn't reachable from user code.

Long CPU-bound loops on the executor (segment merges, decoders, big scans) would otherwise starve other coroutines until they finish. Sprinkling `zio.yield()` works but forces a context switch every time; `maybeYield()` only switches when there's actually contention.

## How

Mirrors `zio.yield()`: fetches the current executor via `getCurrentExecutorOrNull()`, no-ops off-executor, and delegates to `Executor.maybeYield(.reschedule, .allow_cancel)`, returning `Cancelable!void`.

```zig
while (moreWork()) {
    doChunk();
    try zio.maybeYield();
}
```

### DIFF
--- a/src/runtime.zig
+++ b/src/runtime.zig
@@ -681,6 +681,15 @@ pub fn yield() Cancelable!void {
     return task.yield(.reschedule, .allow_cancel);
 }
 
+/// Cooperatively yield, but only if enough other tasks are waiting (a cheap
+/// fairness check for long CPU-bound loops that would otherwise hog the
+/// executor). Returns error.Canceled if the task was canceled. No-op if called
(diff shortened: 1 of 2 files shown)


lalinsky:
@claude add a smoke test for zio.yield and zio.maybeYield

PR #562, opened 2026-07-10
feat: work stealing and improved idling

### LINKED ISSUE #460: Implement work stealing for load balancing across executors

notxorand:
adds work stealing and advanced idle state handling. closes: #460. earlier attempt: #541

## Goal

Idle executors find and run work sitting in other executors' local queues, without losing a task, without a stuck wakeup, and without more than one executor "searching" at a time under normal load.

## Structures

+ **`idle_mask: atomic usize`**: bit `i` set <=> executor `i` is currently parked with no known work.
+ **`searchers: atomic u32`**: 0 or 1. Caps concurrent active searches to one at a time; the thread holding it is responsible for eventually releasing it.

## Steady state (busy executor, every tick)

`checkAboutForWork`, cheapest first strategy:
1. Main task ready? / local ring non-empty?.
2. Overflow non-empty? refill a batch into the ring, recheck.
3. Still nothing? `stealWork()`: scan other executors (skip any with its `idle_mask` bit set), try `run_queue.steal(victim)`. First success wins.

If none of these find anything, the executor goes idle.

## Going idle

`parkAndSearch`
1. **Publish intent**: set own `idle_mask` bit.
2. **Recheck**: we're trying to make sure nothing lands in-between. If work is found now, clear the bit and return.
3. Otherwise block in `loop.run(.once)`.
4. On return, clear own bit and inspect the *previous* value:
   + **Bit was already clear** means some other thread cleared it for us before we did which is a wake signal. We inherited search duty: check ready -> refill overflow -> `stealWork()` -> give up. Every branch that finds nothing forwards, one and only one call releases `searchers` back to 0.
   + **Bit was still set** means we cleared it ourselves; we just woke for unrelated I/O, not because anyone was signaling us. No search obligation.

## Waking someone (called after any push, and after a successful steal)

`armSearcher`
1. CAS `searchers` 0 -> 1. If this fails then someone's already searching, we're done here.
2. Read `idle_mask`. If it's empty, no one to wake, release the token immediately.
3. Pick the lowest set bit, clear it via `fetchAnd` (this clear *is* the token handoff), check the previous value:
   + If the previous bit was already cleared (target was self-cleared concurrently) it means we lost the race, release the token.
   + Otherwise, `wake()` that executor's loop.

**Token handoff**: searcher token is never passed by a separate message — it's implied by *which thread's clear of a given idle-mask bit "wins."* If `armSearcher` clears it, that executor inherits the search obligation on wake. If the executor clears its own bit first (step 4 above), no one ever gets to claim it and there was nothing to hand off. This is why every code path that finds a bit already clear must immediately release `searchers`.

## Two liveness rules

+ **Re-arm on leftover work**: any time a search step (refill, steal) finds work but leaves more visible behind it (steal's victim not empty, overflow not drained), we call `armSearcher()` again before returning not to risk any leftover work having no one looking for it until the next unrelated wake.
+ **Release before final recheck**: when a search is about to give up entirely, release `searchers` *before* the last check, not after so a concurrent pusher racing that exact moment can arm a fresh search instead of finding the token falsely held by a thread that's about to abandon it anyway.

## Other Changes

+ `max_executors` and executor ids (`ExecutorId`) were made architecture scoped to follow u32 on 32-bit arch and u64 on 64-bit arch, with their corresponding bit shifting values (u5 for 32-bit, u6 for 64-bit)

### DIFF
--- a/src/runtime.zig
+++ b/src/runtime.zig
@@ -1172,21 +1172,19 @@ pub const Runtime = struct {
     fn armSearcher(self: *Runtime) void {
         if (!(zio_options.task_migration and self.options.enable_task_migration) or !self.executors_stealable.load(.acquire) or (self.searchers.load(.monotonic) != 0)) return;
         if (self.searchers.cmpxchgStrong(0, 1, .acq_rel, .monotonic)) |_| return;
-        const idle_mask = self.idle_mask.load(.acquire);
 
-        if (idle_mask == 0) {
-            _ = self.searchers.cmpxchgStrong(1, 0, .acq_rel, .monotonic);
-            return;
+        var candidates = self.idle_mask.load(.acquire);
+        while (candidates != 0) {
+            const id: ExecutorId = @intCast(@ctz(candidates));
+            const bit = @as(usize, 1) << id;
+            const previous_mask = self.idle_mask.fetchAnd(~bit, .acq_rel);
+            if (previous_mask & bit != 0) {
+                self.executors.items[id].loop.wake();
+                return;
+            }
+            candidates &= ~bit;
         }
-        const id: ExecutorId = @intCast(@ctz(idle_mask));
-        const bit = @as(usize, 1) << id;
-        const previous_mask = self.idle_mask.fetchAnd(~bit, .acq_rel);
-
-        if (previous_mask & bit == 0) {
-            _ = self.searchers.cmpxchgStrong(1, 0, .acq_rel, .monotonic);
-            return;
-        }
-        self.executors.items[id].loop.wake();
+        _ = self.searchers.cmpxchgStrong(1, 0, .acq_rel, .monotonic);
     }
 
     // Convenience methods that operate on the current coroutine context
--- a/src/runtime.zig
+++ b/src/runtime.zig
@@ -38,7 +38,7 @@ const select = @import("select.zig");
 const Waiter = @import("common.zig").Waiter;
 const random_mod = @import("random.zig");
 
-const U5U6 = switch (@sizeOf(usize)) {
+const ExecutorId = switch (@sizeOf(usize)) {
     4 => u5,
     8 => u6,
     else => @compileError("Unsupported architecture"),
@@ -289,13 +289,9 @@ pub fn getNextExecutor(rt: *Runtime) error{RuntimeShutdown}!*Executor {
 
 // Executor - per-thread execution unit for running coroutines
 pub const Executor = struct {
-    pub const max_executors = switch (@sizeOf(usize)) {
-        4 => 32,
-        8 => 64,
-        else => @compileError("Unsupported architecture"),
-    };
+    pub const max_executors = std.math.maxInt(ExecutorId) + 1;
 
-    id: U5U6,
+    id: ExecutorId,
     loop: ev.Loop,
 
     /// Per-executor random state (non-secure CSPRNG; later the secure-path fd/handle).
@@ -370,7 +366,7 @@ pub const Executor = struct {
         return @alignCast(@fieldParentPtr("main_task", main_task));
     }
 
-    pub fn init(self: *Executor, runtime: *Runtime, id: U5U6) !void {
+    pub fn init(self: *Executor, runtime: *Runtime, id: ExecutorId) !void {
         self.* = .{
             .id = id,
             .loop = undefined,
@@ -1014,7 +1010,7 @@ pub const Runtime = struct {
                 const worker = self.workers.addOneAssumeCapacity();
                 errdefer _ = self.workers.pop();
                 worker.* = .{};
-                worker.thread = try std.Thread.spawn(.{}, runWorker, .{ self, worker, @as(U5U6, @intCast(i + worker_id_start)) });
+                worker.thread = try std.Thread.spawn(.{}, runWorker, .{ self, worker, @as(ExecutorId, @intCast(i + worker_id_start)) });
             }
 
             for (self.workers.items, 0..) |*worker, i| {
@@ -1150,7 +1146,7 @@ pub const Runtime = struct {
 
     /// Worker thread entry point. Initializes executor and runs until stopped.
     /// Signals worker.ready after initialization (success or failure).
-    fn runWorker(self: *Runtime, worker: *Worker, id: U5U6) void {
+    fn runWorker(self: *Runtime, worker: *Worker, id: ExecutorId) void {
         worker.executor.init(self, id) catch |e| {
             worker.err = e;
             worker.ready.set();
@@ -1183,7 +1179,7 @@ pub const Runtime = struct {
             _ = self.searchers.cmpxchgStrong(1, 0, .acq_rel, .monotonic);
             return;
         }
-        const id: U5U6 = @intCast(@ctz(idle_mask));
+        const id: ExecutorId = @intCast(@ctz(idle_mask));
         const bit = @as(usize, 1) << id;
         const previous_mask = self.idle_mask.fetchAnd(~bit, .acq_rel);
 
@@ -1596,11 +1592,11 @@ test "runtime: local ring overflow spills and drains with task migration disable
     try std.testing.expectEqual(@as(u32, H.n_tasks), counter.load(.monotonic));
 }
 
-test "runtime: multi-threaded execution with 64 executors" {
-    const runtime = try Runtime.init(std.testing.allocator, .{ .executors = .exact(64) });
+test "runtime: multi-threaded execution with max executors" {
+    const runtime = try Runtime.init(std.testing.allocator, .{ .executors = .exact(Executor.max_executors) });
     defer runtime.deinit();
 
-    try std.testing.expectEqual(64, runtime.executors.items.len);
+    try std.testing.expectEqual(Executor.max_executors, runtime.executors.items.len);
 }
 
 test "Runtime: multi-threaded with task migration" {
--- a/src/runtime.zig
+++ b/src/runtime.zig
@@ -595,7 +595,7 @@ pub const Executor = struct {
     }
 
     fn stealWork(self: *Executor) bool {
-        if (!zio_options.task_migration) return false;
+        if (!zio_options.task_migration or !self.runtime.executors_stealable) return false;
         const executors = self.runtime.executors.items;
         if (executors.len <= 1) return false;
 
@@ -933,6 +933,7 @@ pub const Runtime = struct {
     options: RuntimeOptions,
 
     executors: std.ArrayList(*Executor) = .empty,
+    executors_stealable: bool = false,
     // Shared global run queue (used when task migration is on): external
     // submissions, cross-thread wakes, and per-executor ring overflow land here,
     // and every executor drains a fair batch from it once per tick.
@@ -1021,6 +1022,7 @@ pub const Runtime = struct {
                 }
                 self.executors.appendAssumeCapacity(&worker.executor);
             }
+            self.executors_stealable = true;
         }
     }
 
@@ -1040,6 +1042,7 @@ pub const Runtime = struct {
             worker.thread.join();
         }
         self.workers.deinit(self.allocator);
+        self.executors_stealable = false;
     }
 
     pub fn deinit(self: *Runtime) void {
--- a/src/runtime.zig
+++ b/src/runtime.zig
@@ -602,7 +602,12 @@ pub const Executor = struct {
             const victim_idle = self.runtime.idle_mask.load(.acquire) & victim_bit != 0;
             if (victim_idle) continue;
 
-            if (self.run_queue.steal(&victim.run_queue)) |node| {
+            if (self.run_queue.steal(&victim.run_queue, struct {
+                fn reset(node: *WaitNode) void {
+                    const task = AnyTask.fromWaitNode(node);
+                    task.last_run_tick = 0;
+                }
+            }.reset)) |node| {
                 self.run_queue.push(node);
                 self.runtime.armSearcher();
                 return true;
--- a/src/runtime.zig
+++ b/src/runtime.zig
@@ -284,7 +284,11 @@ pub fn getNextExecutor(rt: *Runtime) error{RuntimeShutdown}!*Executor {
 
 // Executor - per-thread execution unit for running coroutines
 pub const Executor = struct {
-    pub const max_executors = 64;
+    pub const max_executors = switch (@sizeOf(usize)) {
+        4 => 32,
+        8 => 64,
+        else => @compileError("Unsupported architecture"),
+    };
 
     id: u6,
     loop: ev.Loop,
@@ -520,25 +524,18 @@ pub const Executor = struct {
     }
 
     fn parkAndSearch(self: *Executor, check_ready: bool) !void {
-        const my_bit = @as(u64, 1) << self.id;
-        self.runtime.idle_mask_mutex.lock();
-        self.runtime.idle_mask |= my_bit;
-        self.runtime.idle_mask_mutex.unlock();
+        const my_bit = @as(usize, 1) << self.id;
+        _ = self.runtime.idle_mask.fetchOr(my_bit, .acq_rel);
 
         if (self.checkAboutForWork(check_ready)) {
-            self.runtime.idle_mask_mutex.lock();
-            self.runtime.idle_mask &= ~my_bit;
-            self.runtime.idle_mask_mutex.unlock();
+            _ = self.runtime.idle_mask.fetchAnd(~my_bit, .acq_rel);
             try self.loop.run(.no_wait);
             return;
         }
 
         try self.loop.run(.once);
 
-        self.runtime.idle_mask_mutex.lock();
-        const was_already_clear = self.runtime.idle_mask & my_bit == 0;
-        self.runtime.idle_mask &= ~my_bit;
-        self.runtime.idle_mask_mutex.unlock();
+        const was_already_clear = self.runtime.idle_mask.fetchAnd(~my_bit, .acq_rel) & my_bit == 0;
 
         if (was_already_clear) {
             if (self.main_task.state.load(.acquire).tag == .ready or !self.run_queue.isEmpty()) {
@@ -605,10 +602,8 @@ pub const Executor = struct {
             const victim = executors[(self.id + i) % executors.len];
             if (victim == self) continue;
 
-            const victim_bit = @as(u64, 1) << victim.id;
-            self.runtime.idle_mask_mutex.lock();
-            const victim_idle = self.runtime.idle_mask & victim_bit != 0;
-            self.runtime.idle_mask_mutex.unlock();
+            const victim_bit = @as(usize, 1) << victim.id;
+            const victim_idle = self.runtime.idle_mask.load(.acquire) & victim_bit != 0;
             if (victim_idle) continue;
 
             if (self.run_queue.steal(&victim.run_queue)) |node| {
@@ -941,9 +936,7 @@ pub const Runtime = struct {
     // submissions, cross-thread wakes, and per-executor ring overflow land here,
     // and every executor drains a fair batch from it once per tick.
     global_overflow: OverflowQueue(WaitNode) = .{},
-    idle_mask: u64 = 0,
-    // TODO: replace with atomic scoped based on architecture
-    idle_mask_mutex: OsMutex = .init(),
+    idle_mask: std.atomic.Value(usize) = .init(0),
     searchers: std.atomic.Value(u32) = .init(0),
     loop_group: ev.LoopGroup = .{},
     main_executor: Executor,
@@ -1179,15 +1172,20 @@ pub const Runtime = struct {
 
     fn armSearcher(self: *Runtime) void {
         if (self.searchers.cmpxchgStrong(0, 1, .acq_rel, .monotonic)) |_| return;
-        self.idle_mask_mutex.lock();
-        defer self.idle_mask_mutex.unlock();
-        if (self.idle_mask == 0) {
+        const idle_mask = self.idle_mask.load(.acquire);
+
+        if (idle_mask == 0) {
             _ = self.searchers.cmpxchgStrong(1, 0, .acq_rel, .monotonic);
             return;
         }
-        const id: u6 = @intCast(@ctz(self.idle_mask));
-        const bit = @as(u64, 1) << id;
-        self.idle_mask &= ~bit;
+        const id: u6 = @intCast(@ctz(idle_mask));
+        const bit = @as(usize, 1) << id;
+        const previous_mask = self.idle_mask.fetchAnd(~bit, .acq_rel);
+
+        if (previous_mask & bit == 0) {
+            _ = self.searchers.cmpxchgStrong(1, 0, .acq_rel, .monotonic);
+            return;
+        }
         self.executors.items[id].loop.wake();
     }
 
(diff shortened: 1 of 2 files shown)


notxorand:
@lalinsky what do you think of the sized idle_mask, max_executors and executor id? that or mutex guarding?

src/runtime.zig
```diff
@@ -38,6 +38,11 @@ const select = @import("select.zig");
 const Waiter = @import("common.zig").Waiter;
 const random_mod = @import("random.zig");
 
+const U5U6 = switch (@sizeOf(usize)) {
```
lalinsky: Call this `ExecutorId`

src/runtime.zig
```diff
@@ -284,9 +289,13 @@ pub fn getNextExecutor(rt: *Runtime) error{RuntimeShutdown}!*Executor {
 
 // Executor - per-thread execution unit for running coroutines
 pub const Executor = struct {
-    pub const max_executors = 64;
+    pub const max_executors = switch (@sizeOf(usize)) {
```
lalinsky: You can use `std.math.maxInt(ExecutorId) + 1`

lalinsky (commented):
not a full review yet, but the idle_mask sizing is fine this way

lalinsky:
there are multiple logical issues in this, but please write the design overview how it's supposed to work, so that I don't overwhelm you with comments.

lalinsky:
one partial review, since I was interested in the riscv32 crash, you are reaching out to other executors during Runtime.init, so accessing unitialized executors from the array, stealing needs to be gated to start only after all executors are initialized, and disable it at shutdown.

notxorand:
@lalinsky, I've included a design overview as the pr body info

lalinsky:
I'm sorry, but there are many issues, so to save my time, I'll have AI review this first.

src/runtime.zig
```diff
@@ -505,42 +510,113 @@ pub const Executor = struct {
...
+
+    fn parkAndSearch(self: *Executor, check_ready: bool) !void {
+        const my_bit = @as(usize, 1) << self.id;
+        _ = self.runtime.idle_mask.fetchOr(my_bit, .acq_rel);
+
+        if (self.checkAboutForWork(check_ready)) {
+            _ = self.runtime.idle_mask.fetchAnd(~my_bit, .acq_rel);
+            try self.loop.run(.no_wait);
+            return;
+        }
```
lalinsky: **Searcher token leak, permanently disables all future wakes.** This recheck ignores the result of clearing our own idle bit. If a concurrent `armSearcher` already claimed our bit (it CAS'd `searchers` 0->1, cleared our bit, and woke us), the token is now ours per the handoff protocol, but this path never releases it. `searchers` sticks at 1 forever, every future `armSearcher` fails its CAS and returns, and steal wake-ups silently die for the rest of the runtime's life. There's no self-heal path: the releases below are only reachable via `was_already_clear`, which requires an `armSearcher` success that can no longer happen.

This isn't even a rare cross-thread race: `checkAboutForWork` here calls `stealWork`, which on success calls `armSearcher` *while our own idle bit is set*, and `@ctz` can pick **our own bit** (guaranteed if we're the lowest idle id). So the executor claims the token, wakes itself, and leaks it single-handedly.

The `fetchAnd` result needs the same `was_already_clear` handling as the post-park path, and `armSearcher` probably shouldn't be able to target the calling executor.

src/runtime.zig
```diff
@@ -505,42 +510,113 @@ pub const Executor = struct {
...
+
+        try self.loop.run(.once);
```
lalinsky: If `loop.run(.once)` fails, the `fetchAnd` below is skipped and the worker sits in `runWorker`'s backoff sleep with its idle bit still set, absorbing `armSearcher` wakes (and leaking the token via the was-set path when it re-parks). The bit should be cleared on the error path too (`defer`/`errdefer`).

src/runtime.zig
```diff
@@ -505,42 +510,113 @@ pub const Executor = struct {
...
+
+        const was_already_clear = self.runtime.idle_mask.fetchAnd(~my_bit, .acq_rel) & my_bit == 0;
+
+        if (was_already_clear) {
+            if (self.main_task.state.load(.acquire).tag == .ready or !self.run_queue.isEmpty()) {
```
lalinsky: This ignores `check_ready`, and worker executors' `main_task.state` is initialized `.ready` and never changes (a worker's main task does no I/O). So a woken worker always takes this early-out: release token, do nothing. The whole inherited-search block below (refill / steal / re-arm) is dead code on workers. It works out only because the outer `run()` loop immediately re-runs `checkAboutForWork` (which refills/steals), but that search happens token-less and the protocol as described in the PR never actually executes on workers. This should be `check_ready and ...` like the check in `checkAboutForWork`.

src/runtime.zig
```diff
@@ -505,42 +510,113 @@ pub const Executor = struct {
...
-                // With migration on the overflow is the shared global queue, so
-                // take only a fair ~1/n_exec slice to avoid monopolizing it. With
-                // migration off it is this executor's own private queue and we are
-                // its sole drainer, so take as much as fits (refill caps it).
-                const batch = if (self.runtime.options.enable_task_migration)
-                    pending / @max(self.runtime.executors.items.len, 1) + 1
-                else
-                    pending;
-                self.run_queue.refill(batch);
+                self.run_queue.refill(@min(pending, 64));
```
lalinsky: nit: `refill(@min(pending, 64))` bypasses the fair `pending / n_exec + 1` slice policy used on the same (global) queue in `checkAboutForWork`. Any reason for the different policy here?

src/runtime.zig
```diff
@@ -505,42 +510,113 @@ pub const Executor = struct {
...
-            // Check again after I/O
-            if (check_ready and self.main_task.state.load(.acquire).tag == .ready) {
-                return;
+            _ = self.runtime.searchers.cmpxchgStrong(1, 0, .acq_rel, .monotonic);
+            if (self.stealWork()) return;
+
+            // Release the token before the final recheck. A concurrent
+            // pusher that sees searchers == 0 can arm a fresh search instead of
+            // being blocked behind a token we're about to give up anyway.
+            _ = self.runtime.searchers.cmpxchgStrong(1, 0, .acq_rel, .monotonic);
```
lalinsky: **Double release of the searcher token.** The release here at the top runs before `stealWork`; if the steal fails, the second cmpxchg below "releases" again. If another thread's `armSearcher` armed and handed a token to executor H in between, that second cmpxchg(1->0) destroys H's token, breaking the at-most-one-searcher cap and the stated "one and only one call releases" invariant. Benign for liveness (a stray *clear* allows extra searchers rather than losing wakes), but the protocol should have exactly one release per obligation.

src/runtime.zig
```diff
@@ -505,42 +510,113 @@ pub const Executor = struct {
...
+            if (self.checkAboutForWork(true)) self.runtime.armSearcher();
+        }
+    }
+
+    fn checkAboutForWork(self: *Executor, check_ready: bool) bool {
+        // Run event loop - non-blocking if there's local work, otherwise wait
+        // for I/O. Overflow-queue work needn't be checked here: any cross-thread
+        // push to the overflow queue also wakes this executor's loop, so .once
+        // returns promptly and the next iteration drains it.
```
lalinsky: nit: this comment was copied from the old run loop and is now wrong, `checkAboutForWork` doesn't run the event loop.

src/runtime.zig
```diff
@@ -505,42 +510,113 @@ pub const Executor = struct {
...
+                pending;
+            self.run_queue.refill(batch);
+            has_work = !self.run_queue.isEmpty();
+        }
+        if (!has_work) has_work = self.stealWork();
+        return has_work;
+    }
+
+    fn stealWork(self: *Executor) bool {
+        if (!zio_options.task_migration or !self.runtime.executors_stealable) return false;
```
lalinsky: **`stealWork` ignores the runtime-level `enable_task_migration` option.** This guard only checks the comptime `zio_options.task_migration` and `executors_stealable`. A runtime created with `.enable_task_migration = false` on a migration-enabled build (documented as "tasks are pinned to their home executor", and what selects the per-executor private overflow queue in `Executor.init`) will still steal tasks across executors, breaking the pinning contract.

src/runtime.zig
```diff
@@ -505,42 +510,113 @@ pub const Executor = struct {
...
+        const executors = self.runtime.executors.items;
+        if (executors.len <= 1) return false;
+
+        for (0..executors.len) |i| {
+            const victim = executors[(self.id + i) % executors.len];
```
lalinsky: nit: the scan starts at `i = 0`, which is always `self` and immediately skipped, and the order is deterministic, so all thieves hammer the same victims in the same order. Go and Tokio randomize the start offset; there's a per-executor `random_state` available for this.

src/runtime.zig
```diff
@@ -505,42 +510,113 @@ pub const Executor = struct {
...
+            if (victim == self) continue;
+
+            const victim_bit = @as(usize, 1) << victim.id;
+            const victim_idle = self.runtime.idle_mask.load(.acquire) & victim_bit != 0;
+            if (victim_idle) continue;
+
+            if (self.run_queue.steal(&victim.run_queue)) |node| {
+                const task = AnyTask.fromWaitNode(node);
+                task.last_run_tick = 0;
+                self.run_queue.push(node);
```
lalinsky: **Stolen batch keeps stale `last_run_tick`.** Only the single returned node is reset here; the roughly half-ring that `steal()` published directly into the thief's ring keeps the victim's tick values. If one of them happens to equal the thief's `current_tick`, `isRunnableThisTick` blocks the ring head (and everything behind it) until the next tick. The migrate path in `scheduleTask` resets `last_run_tick` for exactly this reason; the stolen batch needs the same treatment.

src/runtime.zig
```diff
@@ -943,6 +1022,7 @@ pub const Runtime = struct {
                 }
                 self.executors.appendAssumeCapacity(&worker.executor);
             }
+            self.executors_stealable = true;
```
lalinsky: **Data race.** `executors_stealable` is a plain bool written here and read by every worker in `stealWork` with no atomics. Beyond being formally UB, there's no happens-before edge with the `executors.appendAssumeCapacity` calls it's guarding, so a worker on a weakly-ordered machine (ppc64le, riscv, we've been bitten before) can observe `true` but a stale `executors.items.len`. This should be `std.atomic.Value(bool)` with a release store here and an acquire load in `stealWork` (or an atomic executor count).

src/runtime.zig
```diff
@@ -1093,6 +1174,25 @@ pub const Runtime = struct {
...
+
+        if (idle_mask == 0) {
+            _ = self.searchers.cmpxchgStrong(1, 0, .acq_rel, .monotonic);
+            return;
+        }
+        const id: ExecutorId = @intCast(@ctz(idle_mask));
+        const bit = @as(usize, 1) << id;
+        const previous_mask = self.idle_mask.fetchAnd(~bit, .acq_rel);
+
+        if (previous_mask & bit == 0) {
```
lalinsky: nit: on a lost bit claim this gives up entirely instead of retrying with the remaining mask, so a wake opportunity is dropped. Easy to fix with a loop over the mask.

lalinsky (changes requested):
Thanks for the detailed writeup, the idle mask + single searcher token design (Tokio-style) is a reasonable shape, and the description of the token handoff protocol is well thought out. Tests pass here in both `-Dtask-migration` modes. However, there is one design gap that undermines the feature, and a couple of real protocol bugs. Details are in the inline comments; the big one plus a few things that don't anchor to changed lines are below.

### Nothing ever arms a searcher on push, so the steady-state case of #460 is still serial

The description says `armSearcher` is "called after any push, and after a successful steal". The push half doesn't exist in the code. `armSearcher` has exactly three call sites, all inside the search/steal machinery itself (`parkAndSearch` x2, `stealWork`). `scheduleTaskLocal`, `scheduleTaskRemote`, the migrate path in `scheduleTask`, and the ring-to-overflow spill all schedule work without arming anyone.

Consequence: an idle executor parked in `loop.run(.once)` is only ever woken to steal by a *chain* started from another successful steal, and the first steal can only happen when some executor is already awake and out of work. In the motivating scenario, executor A wakes many waiting tasks (they migrate to A's local ring via the migrate path in `scheduleTask`), while B/C/D are parked with no pending I/O. Nothing wakes them, A's ring fills, spills to the global overflow (which also wakes no one), and everything runs serially until an unrelated wake happens to land. That's exactly the situation #460 describes, so as written this mostly adds "steal before parking", not steady-state rebalancing. `scheduleTaskLocal` and/or the spill path need to call `armSearcher` (like Tokio's notify-on-push / Go's `wakep`).

### Other items that don't anchor to changed lines

- `worker_id_start` in `threadsStart` is still `u6`; should be `ExecutorId`.
- The `.new` check in `scheduleTask` with its "can be removed once we have work stealing (#460)" comment is untouched by a PR that claims to close #460. Given the missing push-side arming, removing it now would actually make things worse, but the PR shouldn't claim to close #460 while leaving it.
- `max_executors` silently drops from 64 to 32 on 32-bit targets. That's fine given `idle_mask: usize`, but it's a user-visible change worth noting in the changelog.
- None of the added/updated tests exercise the searcher protocol (the only test change is the max-executors rename). I'd like to see a stress test that asserts parked executors actually get woken to steal, e.g. one executor waking a batch of migrated tasks while the others are parked with no pending I/O.

### Summary

The token-handoff idea is sound, but: the push-side arming from the description was never implemented (so the feature doesn't fire in the steady state), and two paths violate the handoff invariant, one of which permanently and silently disables wake-ups (see the `parkAndSearch` recheck comment). I'd ask for: push-side `armSearcher`, handling the `fetchAnd` result in the pre-park recheck (and preventing self-targeting in `armSearcher`), the `enable_task_migration` runtime guard in `stealWork`, an atomic `executors_stealable`, plus a test for the wake path.

src/runtime.zig
notxorand: can I modify `steal` to handle this?

lalinsky: I'd prefer a comptime fn passes to steal that does that, but given that there are no other uses for the queue, you can also do it directly there.

src/runtime.zig
```diff
@@ -505,42 +510,106 @@ pub const Executor = struct {
...
 
-            // Check again after I/O
-            if (check_ready and self.main_task.state.load(.acquire).tag == .ready) {
-                return;
+            // Release the token before the final recheck. A concurrent
+            // pusher that sees searchers == 0 can arm a fresh search instead of
+            // being blocked behind a token we're about to give up anyway.
+            _ = self.runtime.searchers.cmpxchgStrong(1, 0, .acq_rel, .monotonic);
+            if (self.stealWork()) return;
+            if (self.checkAboutForWork(true)) self.runtime.armSearcher();
```
lalinsky: This is the wake storm described in the review body: `check_ready` is hardcoded `true`, and a worker's `main_task.state` is permanently `.ready`, so an idle worker that just finished a failed search always "finds work" here and arms another searcher, which wakes another idle worker into the same dead end. Self-sustaining chain, measured at ~40-50k spurious wakes/s with 3 idle workers.

Beyond passing `check_ready`: `main_ready` should not be an arm reason at all. Only this executor can run its own main task, so waking a searcher for it is always wasted (that applies to the main executor too). Arm here only for work someone else can take (overflow non-empty / steal leftovers).

src/runtime.zig
```diff
@@ -505,42 +510,106 @@ pub const Executor = struct {
...
+            if (check_ready and self.main_task.state.load(.acquire).tag == .ready) {
+                return;
+            }
+        }
+    }
+
+    fn parkAndSearch(self: *Executor, check_ready: bool) !void {
+        const my_bit = @as(usize, 1) << self.id;
+        _ = self.runtime.idle_mask.fetchOr(my_bit, .acq_rel);
+        errdefer _ = self.runtime.idle_mask.fetchAnd(~my_bit, .acq_rel);
```
lalinsky: Small residue of the token-leak fix: this errdefer discards the `fetchAnd` result. If `armSearcher` claimed our bit just before `loop.run` failed, the token was handed to us and this path never releases it, so `searchers` sticks at 1 permanently (same failure mode as the leak fixed in this revision, just confined to the rare loop-error path). The errdefer should release `searchers` when the previous value shows the bit was already clear.

src/runtime.zig
```diff
@@ -587,6 +656,7 @@ pub const Executor = struct {
         std.debug.assert(getCurrentExecutorOrNull() == self);
 
         self.run_queue.push(&task.awaitable.wait_node);
+        self.runtime.armSearcher();
```
lalinsky, line 669: Hot-path concern: this is the yield/wake fast path, and `armSearcher`'s check order makes the fully-busy steady state (searchers == 0, idle_mask == 0) the most expensive case: every local push does a successful CAS 0->1 on `searchers`, a load of `idle_mask`, and a CAS back 1->0. That's two RMWs on one globally shared cache line per push, serializing all executors under load. Cheap fix: load `idle_mask` first and return if it's 0 (a monotonic load of a mostly-read line, true whenever everyone is busy) before ever touching `searchers`. Tokio does the equivalent (check for idle workers before trying to become the notifier). Has this been benchmarked against main under a busy multi-executor workload?

src/runtime.zig
```diff
@@ -505,42 +510,106 @@ pub const Executor = struct {
...
+            if (victim == self) continue;
+
+            const victim_bit = @as(usize, 1) << victim.id;
+            const victim_idle = self.runtime.idle_mask.load(.acquire) & victim_bit != 0;
+            if (victim_idle) continue;
+
+            if (self.run_queue.steal(&victim.run_queue)) |node| {
+                self.run_queue.push(node);
+                self.runtime.armSearcher();
+                return true;
```
lalinsky: The previous revision reset `last_run_tick` for the returned node; now nothing is reset, while the migrate path in `scheduleTask` still resets it. The ~half-ring that `steal()` publishes into the thief's ring keeps the victim's tick values, and any of them equal to the thief's `current_tick` makes `isRunnableThisTick` block the ring head (and everything behind it) until the next tick. One-tick latency, not correctness, but it should be consistent with the migrate path; resetting the whole stolen batch would resolve both.

src/runtime.zig
```diff
@@ -962,6 +1037,7 @@ pub const Runtime = struct {
             worker.thread.join();
         }
         self.workers.deinit(self.allocator);
+        self.executors_stealable.store(false, .release);
```
lalinsky: Shutdown hardening: consider storing `false` at the *start* of `shutdownWorkers` (before the notify loop) rather than after join. There's a narrow window where `armSearcher` claims a parked worker's idle bit, the worker meanwhile wakes via the shutdown notify, exits `run()`, and `deinit`s its loop before `armSearcher` gets to `loop.wake()`, a use-after-free on the loop. Flipping the flag first stops new arms during shutdown and shrinks that window to nothing for the arm path.

lalinsky (changes requested):
Thanks for the quick turnaround. The new revision resolves most of the previous round properly: push-side arming is in, the token leak on the pre-park recheck is fixed by the restructure (including the self-targeting case), `stealWork`/`armSearcher` now honor the runtime `enable_task_migration` option, `executors_stealable` is a proper atomic, the double release is gone (exactly one release per token path now), the woken-worker path honors `check_ready`, the idle bit is cleared on the error path, refill uses the fair slice, the victim scan is randomized, and `armSearcher` retries candidates. Tests pass here in both `-Dtask-migration` modes.

However, while verifying the fixes I found a serious runtime misbehavior, and I can reproduce it deterministically on this revision.

### Idle wake storm: the final recheck arms a searcher forever

`parkAndSearch`'s final recheck calls `checkAboutForWork(true)` with `check_ready` hardcoded to `true`. A worker executor's `main_task.state` is initialized `.ready` and effectively never leaves it (the worker main task is only the scheduler context), so for an idle worker that just drained a failed search, `checkAboutForWork(true)` returns true via `main_ready` even when there is zero work in the system. It then arms a searcher, which wakes another idle worker, which runs the same failed search, hits the same phantom-ready final recheck, and arms again. The chain is self-sustaining as long as at least two workers are idle; it only dies when the token happens to land on the parked main executor (whose main task is genuinely `.waiting`, so its recheck returns false) or when a candidate race empties the scan.

Repro (4 executors, main spawns bursts of trivial tasks with busy gaps, then busy-loops so it never parks): with counters added to `armSearcher`, **every run shows 70k-125k spurious wakes over a ~5s run (~40-50k/s), 100% of them attributed to the final recheck returning true via `main_ready` on a worker executor** (I instrumented the reason: `main=125212 ring=0 refill=0 steal=0`). That's tens of thousands of wasted wake syscalls, epoll returns, and context switches per second while the pool is idle, sustained indefinitely as long as any scheduling activity keeps re-seeding it. On the previous revision this was mostly masked by the (differently buggy) early-out that never let workers reach the final recheck, though I did catch it storming intermittently there too, likely via the rarer main-task park/wake flip (`AnyTask.yield` with `.park` on a main task re-enters `run(.until_ready)`).

Two things need to change:

1. The final recheck must pass `check_ready`, not `true`.
2. More fundamentally, a ready **main task should never justify arming a searcher**: only this executor can run its own main task, so waking someone else to "search" for it is always a wasted wake (this also applies to the main executor with `check_ready = true`). The arm condition on the final recheck should be based only on work someone else could actually take: overflow non-empty or a successful steal that left leftovers behind.

And this is exactly the kind of thing the missing test coverage from the last round would catch: a test that runs a multi-executor runtime, makes it go idle while the main task stays busy, and asserts the process doesn't keep waking workers (e.g. bound a wake counter, or assert on CPU/wake counts) would have failed loudly here. Please add one.

Repro program for reference:

```zig
fn trivial() void {}

pub fn main(init: std.process.Init) !void {
    var rt = try zio.Runtime.init(init.gpa, .{ .executors = .exact(4) });
    defer rt.deinit();

    var group: zio.Group = .init;
    defer group.cancel();

    var x: u64 = 1;
    for (0..200) |_| {
        for (0..4) |_| try group.spawn(trivial, .{});
        for (0..2_000_000) |j| x = x *% 6364136223846793005 +% j;
        std.mem.doNotOptimizeAway(x);
    }
    try group.wait();

    // main busy, never parks; workers fully idle -> storm sustains
    for (0..2_000_000_000) |j| x = x *% 6364136223846793005 +% j;
    std.mem.doNotOptimizeAway(x);
}
```

lalinsky (approved):
We slightly regressed in some benchmarks, but that can be addressed later. Functionally this seems good as is.

PR #573, opened 2026-07-14
Mutex: barge instead of handing off the lock

lalinsky:
The mutex handed the lock to a specific parked waiter on unlock, so every contended transfer waited for the scheduler to run that exact task — a convoy. This rewrites it as a barging lock: woken waiters compete with running tasks. Tasks still park in the mutex's own wait queue; foreign threads now wait directly on the state word with a futex instead of per-waiter Notify wakeups. Platforms without a raw futex (NetBSD) keep using the queue path.

Contended counter benchmark (4 workers × 100k lock/inc/unlock, old/new binaries interleaved per round, n=12):

| mode | old med | new med |
|---|--:|--:|
| tasks, single-threaded | 8.5 ms | 6.8 ms |
| tasks, multi-threaded | 52.0 ms | 41.3 ms |
| OS threads | 3562 ms | 13.3 ms |

The multi-threaded case has high variance in both versions; that turned out to be work-stealing churn rather than the mutex (separate issue). The OS-thread number is the per-waiter Notify handoff being replaced by a plain futex wait on the state word.

Trade-offs: FIFO ordering is gone (same as glibc/Go/parking_lot); a hot lock holder can starve waiters in principle. The wait queue still supports handing off directly, so eventual-fairness (periodic forced handoff) is a small follow-up if it's ever needed.

Adds foreign-thread and mixed tasks+threads tests; full suite passes (549/549).

### DIFF
--- a/src/os/thread.zig
+++ b/src/os/thread.zig
@@ -56,7 +56,7 @@ pub const WakeCount = enum {
 /// - `wake(ptr, count)`: Wake waiting threads (one or all)
 ///
 /// The implementation is selected at compile time based on the target OS.
-const Futex = switch (builtin.os.tag) {
+pub const Futex = switch (builtin.os.tag) {
     .linux => FutexLinux,
     .windows => FutexWindows,
     .freebsd => FutexFreeBSD,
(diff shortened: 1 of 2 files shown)


lalinsky:
@coderabbitai Measured the spin-before-park suggestion before adopting it: 16 interleaved runs of the contended counter (4 OS threads, 100k locks each), spin done exactly like MutexFutex's 100-iteration pattern.

| | med | min | max | futex calls |
|---|--:|--:|--:|--:|
| with spin | 35.1 ms | 21.4 | 47.5 | 361 |
| without | 13.1 ms | 12.0 | 58.8 | 581 |

The spin does what it promises — parking drops by ~40% — and still loses by ~2.7x. With critical sections this short, spinners keep the state word's cache line bouncing between cores, which slows down the holder's unlock and the next acquire more than the saved syscalls are worth; parking takes waiters off the line. Same effect that led to spinning being removed from std.Thread.Mutex. So lockThread stays spin-free.

The cancellation test suggestion was a good catch: added in a28e5f9.

PR #577, opened 2026-07-15
Rewrite RwLock in the shape std.Io.RwLock uses

lalinsky:
The monitor RwLock had two structural problems: one condition shared by reader and writer waiters forced `broadcast()` on every unlock (each departing reader woke every waiter so the last one could hand over), and every operation — including uncontended `lockShared` — went through the internal mutex, serializing the exact workload an RwLock is supposed to parallelize.

This ports the standard library's design onto zio's primitives: atomic state word with reader/writer counts, uncontended shared locking is one CAS, blocked readers queue on the mutex behind the writer, and a writer waiting for readers parks on a semaphore posted by the last reader. Every wakeup is productive by construction.

RwLock churn (8 readers × 20k, 1 writer × 2k, old/new interleaved per round, n=10, medians):

| mode | old | new |
|---|--:|--:|
| pure readers, ST | 7.5 ms | 2.7 ms |
| mixed, ST | 7.9 ms | 3.0 ms |
| pure readers, MT | 92.7 ms | 10.3 ms |
| mixed, MT | 53.8 ms | 10.7 ms |

The MT spread also collapsed (51–106 ms → 8–12 ms) — that's the broadcast herd going away, not just the fast path.

Independent of #573 (works on either mutex; tested on both). The two tests that inspected `writers_waiting`/the shared condition now build the same scenarios through the public API and the state word, including the last-reader-wakes-writer regression. Full suite passes.

Repro: `rwlock_bench_native` in zio-benchmarks (`--pure`, `--zio-mt`).

### DIFF
--- a/src/sync/RwLock.zig
+++ b/src/sync/RwLock.zig
@@ -89,7 +89,15 @@ pub fn lock(self: *RwLock) Cancelable!void {
     const state = @atomicRmw(usize, &self.state, .Add, is_writing -% writer, .seq_cst);
     if (state & reader_mask != 0) {
         self.semaphore.wait() catch |err| {
-            self.unlock();
+            // The final reader may post right as the wait is canceled. Clear
+            // is_writing first; if the readers are already gone, that post is
+            // guaranteed and has to be consumed here, or a later writer would
+            // take it while readers are active.
+            const prev = @atomicRmw(usize, &self.state, .And, ~is_writing, .seq_cst);
+            if (prev & reader_mask == 0) {
+                self.semaphore.waitUncancelable();
+            }
+            self.mutex.unlock();
             return err;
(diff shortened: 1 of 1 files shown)


lalinsky:
@coderabbitai Confirmed and fixed in 3b392a8 — the regression test reliably reproduced the stale permit against the unfixed code ("expected 0, found 1" within a few iterations) and passes with the fix.

Good catch: this is inherited from upstream std.Io.RwLock, which has the same cancel path. Interestingly it's entangled with ziglang/zig#35564 — with the current (unfixed) upstream Condition, the cancel gets swallowed instead of the permit leaking, which is that bug; once #35564 lands, this leak becomes reachable upstream too.

PR #580, opened 2026-07-15
Add a next-up slot to the run queue (Go's runnext)

lalinsky:
Builds on #579. A task woken by the currently running task goes into a next-up slot on the run queue instead of the ring, and runs right after the waker on the same executor.

The motivation is communicating-task placement: a channel ping-pong pair on the multi-threaded runtime used to scatter across executors — the wake pushed the task to the ring and armed a searcher, which promptly stole it, so every message paid a cross-executor wake. The slot placement itself is deliberately not announced; a previous occupant demoted to the ring is ordinary queued work and is announced as usual. The slot is stealable exactly like Go's runnext: a thief that finds the ring empty backs off ~3us to give the owner a chance to run the task itself, then CASes the slot, so a stalled owner can't sit on a runnable task.

Fairness is the same story as Go's inheritTime: a slot chain shares the executor's tick budget (#579), so it gets cut off at the ~100us poll target like any other work, and demoted slot occupants land in the ring where stealing applies.

Wakes from scheduler context (I/O completions, timers) keep going to the ring: they're already local to the right executor and have no waker to pair with.

## Benchmarks

ReleaseFast, medians, same machine as #579.

| benchmark (native API, mt) | before | after | st reference |
|---|---|---|---|
| channel ping-pong | 12.9 ms | **9.2 ms** | 9.0 ms |
| worker pool 1 -> 1000, work=64 | ~2.0M items/s | ~1.8M items/s (noise) | 4.1M |
| worker pool 1000 -> 1 (fan-in) | ~2.5M items/s | ~2.4M items/s | 19M |
| tcp echo 1000 conns | ~225k msgs/s | ~220k msgs/s | 72k |
| tcp echo 1 conn | 41k msgs/s | 41k msgs/s | 55k |

The multi-threaded channel ping-pong now matches the single-threaded runtime — the pair settles on one executor and stays there, since ping-pong wakes never displace a slot occupant. Fan-out/fan-in patterns wake many tasks back to back, so each wake demotes the previous one to the ring with a normal announcement, and their behavior is unchanged.

### DIFF
--- a/src/runtime.zig
+++ b/src/runtime.zig
@@ -744,6 +744,23 @@ pub const Executor = struct {
         self.runtime.armSearcher();
     }
 
+    /// Schedule a task woken by the currently running task: it goes to the run
+    /// queue's next-up slot (Go's runnext) and runs here right after the
+    /// waker, keeping communicating tasks on one executor. The slot placement
+    /// itself is not announced — inviting a thief at that moment would just
+    /// scatter the pair, and a stalled slot can still be stolen (with backoff)
+    /// through the normal steal path. A previous occupant demoted to the ring
+    /// is ordinary queued work, though, and must be advertised.
+    fn scheduleTaskNext(self: *Executor, task: *AnyTask) void {
+        if (task == &self.main_task) return;
+
+        std.debug.assert(getCurrentExecutorOrNull() == self);
+
+        if (self.run_queue.pushNext(&task.awaitable.wait_node)) {
+            self.runtime.armSearcher();
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


lalinsky:
Closing: against the post-#579 baseline this trades ~3ms on channel ping-pong for 20-60% regressions on fan-out/fan-in patterns — a woken task trapped in the slot behind a wake-generator can only be rescued by a thief paying the 3us backoff per task, instead of batch-stealing from the ring. Announce-always (Go's ready/wakep semantics) didn't recover it. Might revisit with a park-aware variant that only slots when the waker is about to block.

PR #581, opened 2026-07-15
I/O load shedding: rebalance sticky I/O across executors

lalinsky:
Found while building the driver-based TCP benchmarks (separate C load generator, driver and server pinned to disjoint cores, so the runtime under test is only ever the server side).

I/O is sticky: a task's ops live on the ring of the executor it runs on — by preference on io_uring, by force on epoll/kqueue. Work stealing cannot rebalance I/O-bound tasks, because their wakes are always local and the wake-to-pop window is sub-microsecond: an idle executor gets searcher-woken tens of thousands of times a second and never catches anything. A multi-threaded TCP server ended up doing all of its work on one core (second executor at 0.0% CPU), byte-for-byte identical to single-threaded on the streaming benchmark.

Go avoids this problem structurally — one shared netpoller means readiness is global and balance is free (at the price of a central funnel, which is why zio beats go on 1000-connection echo). With per-executor rings we need an explicit balancer:

- **Round-robin homes for new tasks** (reverts spawn-onto-current-executor): initial spread is the only placement decision that exists on epoll/kqueue backends, and spawning everything onto the acceptor's executor concentrated whole servers onto one core.
- **Decentralized load shedding**: once per tick (~100us) each executor compares its in-flight I/O count (== tasks parked on its ring) against the runtime average; if above, it sheds the excess by routing up to 4 wakes to the global queue. Less loaded executors pull from it, and a shed task's I/O re-homes wherever it runs next, so one shed rebalances a connection permanently. No central coordinator, no idle-mask races; the cap keeps the global queue a rebalancing channel rather than a scheduling path.

## Benchmarks

Driver/server split, i7-6700 (2 driver cores, 2 server cores), 8 connections, 64KB chunks, ReleaseFast:

| server workload | before | after | go | tokio |
|---|---|---|---|---|
| stream write, 8 conns | 3.5 GB/s | **7.3 GB/s** | 7.1 | 7.4 |
| stream write, 1 conn | 2.7 GB/s | **3.3 GB/s** | 3.8 | 3.6 |
| stream read, 8 conns | 5.0 GB/s | **7.7 GB/s** | 7.5 | 7.7 |
| echo 1000 conns | 146k msgs/s | **166k msgs/s** | 108k | 150k |
| echo 1 conn 4KB | 25.1k msgs/s | 25.7k msgs/s | 23.4k | 24.7k |

In-process benchmarks (channel ping-pong, worker pool fan-out/fan-in, spawn) unchanged within noise — non-I/O workloads keep inflight counts below the shedding threshold, so the balancer never engages there.

Follow-up commit: the first wake pushed onto an empty ring from scheduler context no longer arms a searcher — the executor is in its run loop and pops it immediately, so the announcement only caused an idle executor to occasionally win the race and drag the connection to another ring. That fixed single-connection streaming on mt (2.7 -> 3.3 GB/s, now matching st) and, as a bonus, removed enough wakeup churn to lift 1000-connection echo from 147k to 166k msgs/s.

### DIFF
--- a/src/ev/loop.zig
+++ b/src/ev/loop.zig
@@ -227,6 +227,8 @@ pub const LoopState = struct {
 
     active: usize = 0,
     /// I/O operations submitted to backend awaiting completion
+    // Plain counter mutated only by the owner thread, but read cross-thread
+    // by the scheduler's load shedding, hence the atomic accessors below.
     inflight_io: usize = 0,
 
     wake_requested: std.atomic.Value(u32) = std.atomic.Value(u32).init(0),
@@ -280,12 +282,13 @@ pub const LoopState = struct {
 
     /// Increment the inflight I/O counter. On multi-threaded backends
     /// (IOCP) this routes to a shared atomic; on single-threaded backends
-    /// it's a simple local increment.
+    /// only the owner thread mutates, but other executors read the count for
+    /// load shedding, so the access is a monotonic atomic either way.
     pub fn incrInflight(self: *LoopState) void {
         if (comptime Backend.capabilities.is_multi_threaded) {
             _ = self.loop.loop_group.shared.inflight_io.fetchAdd(1, .monotonic);
         } else {
-            self.inflight_io += 1;
+            @atomicStore(usize, &self.inflight_io, self.inflight_io + 1, .monotonic);
         }
     }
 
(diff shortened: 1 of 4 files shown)


lalinsky:
Reverted the per-executor spawn queue commits (4b482d2, 6a1228b) from this PR — CI still hangs on 2-core runners despite the deferral fixes, and it shouldn't block the load-shedding work. The spawn queue idea and everything learned (callback re-add recursion, defer_callback design, the two stranding bugs) moved to #582; WIP preserved on spawn-queue-wip.

Issue #585, opened 2026-07-16
Memory leak with thread sanitizer?


psznm:
It seems that when zio is run under thread sanitizer, it does not free up memory. 

I have a setup where I run 25 tcp connections simultaenously in one process containing all client, server, and proxy. When I run this for around 40k connections it freezes my computer. In top I can see memory accumulating. When run without thread sanitizer it seems to be able to run indefinitely

EDIT: Actually even without thread sanitizer I seem to be gaining memory usage, even though I do 0 allocations myself past "heating" buffer pool, and even though the count of tasks I have running is not increasing. Gaining memory usage very slowly, but consistently.

psznm:
Is it my fault because I call async or concurrent from within another task and I do that recursively? Intuitively that should be legal, but I am out of ideas.

lalinsky:
Can you show me the pattern in pseudo code?

You mean you are forever chaining more and more spawned tasks? If so, why would you do that?

psznm:
```
fn accept() {
    const stream  = server.accept();
    self.group.async(accept);
    process(stream);
}
```

- my client only does 25 connections at once, so there should usually be 1 accepting task in proxy and in server, and at most 25 processing tasks in each.

lalinsky:
Who owns and awaits the group? Do you have an estimate of the size of the leak? Based on that I could have a guess where is it from.

Btw, that is a pattern that can deadlock you, if you use group.async, you might as well be calling the function directly, so you never get to actually processing the stream, just recurse infinitely. That should be irrelevant to the memory leak, assuming you are not waiting on the group in the accept fn.

psznm:
> Who owns and awaits the group? Do you have an estimate of the size of the leak? Based on that I could have a guess where is it from.

htop is showing me RES to be ~40M seconds from startup and ~400M after 1M requests. That is without thread sanitizer, that is what you were asking about, right?

> Btw, that is a pattern that can deadlock you, if you use group.async, you might as well be calling the function directly, so you never get to actually processing the stream, just recurse infinitely. That should be irrelevant to the memory leak, assuming you are not waiting on the group in the accept fn.

Well the reality is much more complicated than the pseudo code. I am past the deadlocks at this point (at least until I implement fuzzing I guess). I actually do use concurrent - I will be having many accept tasks and this mechanism is intended do "replenish" the tasks after accepting. I could use a queue instead if this is an issue.

lalinsky:
[...]

PR #595, opened 2026-07-21
Always free Tsan fibers

psznm:
Partially solves #585

It does not fully solve the leak, but the memory usage growth is now much milder

```
~/git/zig-playground$ ./run.sh
RSS 22:01:29 = 20188 KB
RSS 22:01:31 = 75140 KB
RSS 22:01:33 = 95284 KB
RSS 22:01:35 = 125088 KB
```

(125M in 6 seconds instead of original 5G)

I couldn't find where the remaining leak is coming from, but I couldn't figure out if it is due to zio's tsan handling (because when tsan handling is disabled, tsan just crashes the example program), or if it just a TSan thing.  Fibers now seem to be freed correctly and zio does nothing else with tsan.

Also this does not at all touch the issue mentioned in [comment](https://github.com/lalinsky/zio/issues/585#issuecomment-5035862059) where TSan caused segfault after destroying seemingly already destroyed fiber, and that issue still persists.

Also as a side effect of this change, weird pause during program exit has disappeared in my program when run with thread sanitizer.

### DIFF
--- a/src/runtime.zig
+++ b/src/runtime.zig
@@ -920,10 +920,7 @@ pub const Executor = struct {
             .finish => |task| {
                 self.pending_cleanup = .none;
                 task.state.store(.{ .tag = .finished }, .release);
-                if (task.coro.context.stack_info.allocation_len > 0) {
-                    self.runtime.stack_pool.release(task.coro.context.stack_info, self.loop.now());
-                    task.coro.context.stack_info.allocation_len = 0;
-                }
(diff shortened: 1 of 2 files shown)


lalinsky:
Oh, thank you. I messed up, the goal was to have the tsan fiber destroy done in two places, but I apparently forgot the main one. I've pushed a commit to your branch with a small refactoring, to destroy the tsan fiber immediately after the task exits, not after it's "reaped".

PR #601, opened 2026-07-22
Treat ENXIO and EOVERFLOW as Unseekable

lalinsky:
Writing to a terminal on macOS crashed with `unexpected error: .NXIO`.

A reader/writer starts in positional mode and expects the first `pread`/`pwrite` to fail with `ESPIPE` when the file is not seekable, which is how it learns to switch to streaming. macOS returns `ENXIO` for a tty instead, so the fallback never ran and the error escaped as `error.Unexpected`.

Map `ENXIO` and `EOVERFLOW` to `error.Unseekable`, same grouping `std.Io.Threaded` uses in its positional paths.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+- Fixed writing to a terminal on macOS crashing with an unexpected `ENXIO` error. A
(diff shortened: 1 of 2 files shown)


lalinsky:
@floatdrop this should fix stdout on macos, I'm surprised we didn't hit this before, most people just use zio with networking only

PR #605, opened 2026-07-23
feat: kqueue native implementation via freebsd sendfile

elect0:
## What
Implements native `sendfile(2)` support for the `kqueue` backend (**FreeBSD/DragonflyBSD**), replacing the generic read/write fallback loop.

## Why
To enable zero-copy network I/O for file serving, replacing the thread-pool + double-buffer fallback, which is slow and CPU-heavy.

## Design
* **Zero-Copy Transfer**: Invokes `sendfile` directly, transferring data straight from the kernel's page cache to the TCP socket buffer, completely bypassing user-space.
* **Cancellation Machinery**: Wires the `.net_send_file` operation into `kqueue.zig` (`getIdent` and `getFilter`). The loop can now correctly identify the socket FD and disarm the `EVFILT.WRITE` event when a task is canceled, resolving the previous `unreachable`.
* **Bounds Checking**: Uses `@min()` to cap the requested transfer size to the actual file size, preventing integer overflows and out-of-bounds sends.

## Test
* Benchmarked a single file transfer via monotonic clock (`.boot`):
  * **Previous (read/write loop):** ~47,500 ns
  * **New (native sendfile):** ~10,000 ns (~4.75x speedup)
* All tests pass.

## Scope
This PR is specifically confined to the `kqueue` backend for FreeBSD/DragonflyBSD platforms.

### DIFF
--- a/src/ev/backends/kqueue.zig
+++ b/src/ev/backends/kqueue.zig
@@ -71,7 +71,7 @@ pub const NetShutdownError = error{
 const Self = @This();
 
 pub const NetSendFileData = struct {
-    offset: usize = 0,
+    offset: u64 = 0,
     remaining: usize = 0,
     // Total bytes sent so far
     sbytes: usize = 0,
@@ -553,10 +553,15 @@ pub fn submit(self: *Self, state: *LoopState, c: *Completion) void {
             const data = c.cast(NetSendFile);
 
             if(data.internal.sbytes == 0 and data.internal.offset == 0 and data.internal.remaining == 0) {
-                const file_size = fs.fileSize(data.file) catch |err| switch (err) {
-                    error.PermissionDenied => return c.setError(error.AccessDenied),
-                    else => |e| return c.setError(e),
+                const file_size = fs.fileSize(data.file) catch |err| {
+                    switch(err) { 
+                        error.PermissionDenied => c.setError(error.AccessDenied),
+                        else => |e| c.setError(e),
+                    }
+                    state.markCompletedFromBackend(c);
+                    return;
                 };
+
                 const avail: u64 = if (data.offset >= file_size) 0 else file_size - data.offset;
                 data.internal.remaining = @min(@as(u64, data.remaining), avail);
                 data.internal.offset = data.offset;
@@ -583,7 +588,9 @@ pub fn submit(self: *Self, state: *LoopState, c: *Completion) void {
                     return;
                 }
 
-                return c.setError(unexpectedError(err));
+                c.setError(unexpectedError(err));
+                state.markCompletedFromBackend(c);                
+                return;
             }
             
             if(data.internal.remaining > 0) {
--- a/src/ev/backends/kqueue.zig
+++ b/src/ev/backends/kqueue.zig
@@ -19,6 +19,7 @@ const NetRecvFrom = @import("../completion.zig").NetRecvFrom;
 const NetSendTo = @import("../completion.zig").NetSendTo;
 const NetRecvMsg = @import("../completion.zig").NetRecvMsg;
 const NetSendMsg = @import("../completion.zig").NetSendMsg;
+const NetSendFile = @import("../completion.zig").NetSendFile;
 const NetPoll = @import("../completion.zig").NetPoll;
 const NetClose = @import("../completion.zig").NetClose;
 const PipePoll = @import("../completion.zig").PipePoll;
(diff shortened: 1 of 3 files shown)


lalinsky:
Thank you. I've resolved the conflict, so that I can allow CI to run. Will review this tomorrow.

src/ev/backends/kqueue.zig
```diff
@@ -550,7 +561,58 @@ pub fn submit(self: *Self, state: *LoopState, c: *Completion) void {
         // File operations are handled by Loop via thread pool
         .file_open, .file_create, .file_close, .file_read, .file_write, .file_sync, .file_size, .file_set_size, .file_set_permissions, .file_set_owner, .file_set_timestamps, .file_stat, .dir_open, .dir_close, .dir_read, .dir_create_dir, .dir_rename, .dir_rename_preserve, .dir_delete_file, .dir_delete_dir, .dir_set_permissions, .dir_set_owner, .dir_set_file_permissions, .dir_set_file_owner, .dir_set_file_timestamps, .dir_sym_link, .dir_read_link, .dir_hard_link, .dir_access, .dir_real_path, .dir_real_path_file, .file_real_path, .file_hard_link, .device_io_control => unreachable,
         // Driven by Loop's generic read/write fallback, never reaches the backend.
-        .net_send_file => unreachable,
+        .net_send_file => {      
+            const data = c.cast(NetSendFile);
+
+            if(data.internal.sbytes == 0 and data.internal.offset == 0 and data.internal.remaining == 0) {
```
lalinsky: You will need to add something like `if (not freebsd) unreachable;` in front of this, otherwise it won't compile on other BSDs.

src/ev/backends/kqueue.zig
```diff
@@ -39,6 +40,7 @@ pub const capabilities: BackendCapabilities = .{
     // The BSDs' EVFILT_TIMER absolute clock is monotonic-only and underspecified
     // with no CLOCK_REALTIME timer, so they keep the capped poll-timeout fallback.
     .native_wall_timers = builtin.os.tag.isDarwin(),
+    .net_send_file = builtin.os.tag == .freebsd or builtin.os.tag == .dragonfly,
```
lalinsky: Dragonfly actually doesn't have `std.c.sendfile` definition in the stdlib. Let's skip it for now and only support FreeBSD.

src/ev/backends/kqueue.zig
```diff
@@ -550,7 +561,58 @@ pub fn submit(self: *Self, state: *LoopState, c: *Completion) void {
...
+
+            var sbytes: std.posix.off_t = 0;
+            const rc = std.c.sendfile(data.file, data.handle, @intCast(data.internal.offset), @intCast(data.internal.remaining), null, &sbytes, 0);
+
+            const sent: usize = @intCast(sbytes);
+            data.internal.offset += sent;
+            data.internal.remaining -= sent;
+            data.internal.sbytes += sent;
+            
+            if(rc == -1) {
```
lalinsky: Please use `zig fmt` on the sources.

src/ev/backends/kqueue.zig
```diff
@@ -550,7 +561,58 @@ pub fn submit(self: *Self, state: *LoopState, c: *Completion) void {
...
+                const err = std.posix.errno(rc);
+                if(err == .AGAIN) {
+                    _ = registerSocket(self, data.handle, .write, false);
```
lalinsky: I'm afraid this will need to be a bit more complex. The `registerSocket` function you are calling is here is part of the cross-loop socket registration and can't be use standalone like this. As is, it would actually hang if you hit `EAGAIN`. I'm not sure if it's better to go with the global socket registration process, or do individual poll for this. If you want, I can finish this part, plus add a test for the large file case that triggers `EAGAIN`.

lalinsky (commented):
Thank you for this. It's a good start, but we need two things:

1) drop dragonfly support

2) fix how we register the poll

src/ev/backends/kqueue.zig
lalinsky: btw, maybe you could open a PR for zig to add the dragonfly sendfile extern.

src/ev/backends/kqueue.zig
elect0: thanks for the review and i would really appreciate if you could take over `EAGAIN` handling part. @lalinsky

PR #608, opened 2026-07-24
fix: pass real environment to Threaded io for process spawning

beckend:
Process spawn fails with FileNotFound for PATH-dependent binaries

processSpawnImpl in src/io.zig created a temporary Io.Threaded with default InitOptions, where environ defaults to .empty. This caused std.process.spawn(rt.io(), ...) to fail with FileNotFound for any binary requiring PATH resolution (e.g. "true", "false", "sleep"), because the Threaded io had no PATH variable to look up.

Root Cause
Io.Threaded.InitOptions.environ defaults to .empty ([`std/Io/Threaded.zig:1600`](https://github.com/...)). With no PATH, the fallback default_PATH (/usr/local/bin:/bin:/usr/bin) misses binaries on NixOS, Homebrew, and other non-standard layouts.

### DIFF
--- a/src/ev/test/process_wait.zig
+++ b/src/ev/test/process_wait.zig
@@ -17,7 +17,7 @@ else
 const argv_kill_self: []const []const u8 = if (builtin.os.tag == .windows)
     &.{}
 else
-    &.{ "sh", "-c", "kill -TERM $$" };
+    &.{ "sh", "-c", "kill -KILL $$" };
 
 test "ProcessWait: wait for child process exit code 0" {
     if (builtin.os.tag == .windows) return error.SkipZigTest;
@@ -40,8 +40,8 @@ test "ProcessWait: wait for child process exit code 0" {
 
     const result = try wait.getResult();
     child.id = null;
-    try std.testing.expectEqual(@as(u8, 0), result.code);
-    try std.testing.expectEqual(@as(?u8, null), result.signal);
+    try std.testing.expectEqual(0, result.code);
+    try std.testing.expectEqual(null, result.signal);
 }
 
 test "ProcessWait: wait for child process exit code 1" {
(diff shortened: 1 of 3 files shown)


src/ev/test/process_wait.zig
```diff
@@ -4,14 +4,91 @@ const Loop = @import("../loop.zig").Loop;
...
+    var child = try std.process.spawn(io, .{ .argv = argv_exit0 });
+    defer if (child.id != null) child.kill(io);
+
+    var wait = ProcessWait.init(child.id.?);
+    loop.add(&wait.c);
+    try loop.run(.until_done);
+
+    const result = try wait.getResult();
+    child.id = null;
+    try std.testing.expectEqual(@as(u8, 0), result.code);
```
lalinsky: You don't need these type casts:

```
try std.testing.expectEqual(0, result.code);
try std.testing.expectEqual(null, result.signal);
```

src/ev/test/process_wait.zig
```diff
@@ -4,14 +4,91 @@ const Loop = @import("../loop.zig").Loop;
...
+
+const argv_exit1: []const []const u8 = if (builtin.os.tag == .windows)
+    &.{ "cmd.exe", "/c", "exit 1" }
+else
+    &.{ "sh", "-c", "exit 1" };
+
+const argv_kill_self: []const []const u8 = if (builtin.os.tag == .windows)
+    &.{}
+else
+    &.{ "sh", "-c", "kill -TERM $$" };
```
lalinsky: It seems that there is a difference in how `sh` deals with this across platforms, and the OpenBSD one will exit with rc 0 on SIGTERM. Maybe we could try using SIGKILL which should be consistent on all platforms.

PR #641, opened 2026-08-01
process: reliably kill children on NetBSD

lalinsky:
## Summary

Make `childKill` reliable on NetBSD while leaving other platforms unchanged:

- Send `SIGKILL` on NetBSD, wait 10ms, and send it again after the vulnerable exec window.
- Route NetBSD process waits through the blocking-pool `waitpid` path instead of `EVFILT_PROC`.
- Other POSIX platforms retain `SIGTERM`; Windows is unchanged.

## Why

NetBSD 10.1 CI consistently took the full `sleep 100` duration even though `kill(pid, SIGTERM)` returned success. Diagnostic runs established two NetBSD kernel issues:

1. [kern/58091](https://gnats.netbsd.org/58091): a signal sent immediately after spawn can be discarded while the child is completing `execve`. This can affect the first SIGKILL too, so the NetBSD workaround retries after 10ms.
2. [kern/60358](https://gnats.netbsd.org/60358): `EVFILT_PROC` cannot reliably attach after the child has already exited. Delegating process waits to blocking `waitpid` avoids that registration race.

Both issues are NetBSD-specific. Temporary tracing and stress tests were removed from the final diff.

## Verification

NetBSD 10.1 VM CI passes in debug and ReleaseSafe for both kqueue and poll. The formerly 100-second test now takes:

- Debug kqueue: 22.22ms
- Debug poll: 20.89ms
- ReleaseSafe kqueue: 15.27ms
- ReleaseSafe poll: 17.24ms

Local verification:

- `./check.sh --filter "childKill: terminates process" --backend poll`
- NetBSD kqueue cross-build with the same test filter

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,10 +12,22 @@ jobs:
       fail-fast: false
       matrix:
         zig_version: ['0.16.0']
-        # Diagnostic PR: iterate quickly on the NetBSD childKill investigation.
-        mode: [debug]
+        mode: [debug, release]
         config:
+          - { os: ubuntu-24.04, target: native, backends: 'io_uring epoll poll' }
+          - { os: ubuntu-24.04-arm, target: native, backends: 'io_uring epoll poll' }
+          - { os: macos-latest, target: native, backends: 'kqueue poll' }
+          - { os: macos-15-intel, target: native, backends: 'kqueue poll' }
+          - { os: windows-latest, target: native, backends: 'iocp poll' }
+          - { os: ubuntu-latest, target: arm-linux, qemu: true, backends: 'epoll poll' }
+          - { os: ubuntu-latest, target: thumb-linux, qemu: true, backends: 'epoll poll' }
+          - { os: ubuntu-latest, target: riscv32-linux, qemu: true, backends: 'epoll poll' }
+          - { os: ubuntu-latest, target: riscv64-linux, qemu: true, backends: 'epoll poll' }
+          - { os: ubuntu-latest, target: loongarch64-linux, qemu: true, backends: 'epoll poll' }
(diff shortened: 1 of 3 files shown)


lalinsky:
Intermediate result from NetBSD 10.1 CI:

- Parent SIGTERM disposition is default and SIGTERM is unblocked.
- `kill(pid, SIGTERM)` returns 0 / SUCCESS.
- With the original immediate spawn-then-kill sequence, the child exits normally after the full 100 seconds (`EVFILT_PROC` data 0), not from SIGTERM.
- A 100-iteration stress test with `sleep 1` produced both normal exits after ~1 second and signal-15 exits, proving a timing-dependent loss before process-wait handling. Poll shows the same behavior, so kqueue is not the primary cause.
- This exactly matches NetBSD kern/58091: after fork/execve or posix_spawn, parent `kill(child, SIGTERM)` is unreliable because execve incorrectly flushes pending signals. The kernel fix landed in March 2025; CI runs NetBSD 10.1, released before that fix.
- Registering ProcessWait after a killed child exits also reproduces the separate EVFILT_PROC/ESRCH behavior (`EV_ERROR`, data 3), matching kern/60358, but waitpid reaps immediately and this does not cause the 100-second delay.

Final experiment now running: compare immediate SIGKILL with a delayed second SIGTERM to identify reliable application-level workarounds on affected NetBSD releases.

lalinsky:
Final workaround run passed. Immediate SIGKILL completed 20/20 times on each backend in about 27ms total; issuing SIGTERM again after 10ms completed 20/20 times on each backend in about 392ms total. The separate EVFILT_PROC registration consistently returned EV_ERROR/ESRCH after prompt termination and waitpid reaped successfully. Root cause and experiment results are now summarized in the PR description.

PR #647, opened 2026-08-02
net: split address and port reuse options

lalinsky:
## Summary

- Add independent native `reuse_address` and `reuse_port` options for IP bind/listen APIs.
- Remove the combined `Socket.setReuse` helper.
- Preserve the existing `std.Io` reuse behavior in its adapter layer.
- Add Linux coverage for address-only versus port reuse listener behavior.

## Validation

- `zig build test`
- 604 tests passed, 1 skipped


lalinsky:
@fixnet-ai this should fix your issue from https://github.com/fixnet-ai/utm-monitor/pull/5

PR #664, opened 2026-08-15
Update net_send/net_write for latest Zig 0.17

zacoons:
I was hoping this would be an easy change so I could test out zio in my DNS proxy, but it looks like implementing this will be complicated. So I figured I'd post what I've done so far in case you find it useful. Broken by https://codeberg.org/ziglang/zig/pulls/35292

### DIFF
--- a/src/io.zig
+++ b/src/io.zig
@@ -1941,9 +1941,10 @@ fn bindErrToListenErr(err: BindOrCancel) Io.net.IpAddress.ListenError {
         error.NetworkDown => error.NetworkDown,
         error.SystemResources => error.SystemResources,
         error.Canceled => error.Canceled,
-        // TODO(zig-0.17): map `error.AccessDenied` through once `ListenError` has it
-        // (https://codeberg.org/ziglang/zig/pulls/36307). Binding a privileged port
-        // without the permission to do so is a normal error, not something unexpected.
+        // TODO: map `error.AccessDenied` through once `ListenError` has it. Binding a
+        // privileged port without the permission to do so is a normal error, not
+        // something unexpected. Still missing from the set as of 0.17.0-dev.1464,
+        // waiting on https://codeberg.org/ziglang/zig/pulls/36307.
         error.AccessDenied,
         error.FileDescriptorNotASocket,
         error.SymLinkLoop,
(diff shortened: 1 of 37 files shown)


src/io.zig
```diff
@@ -532,18 +545,26 @@ const BatchCompletionData = union(Io.Operation.Tag) {
...
     },
+    net_send: struct {
+        op: ev.NetSendMsg,
+    },
     net_read: struct {
         op: ev.NetRecv,
         iovecs: [max_iovecs_len]os_net.iovec,
     },
+    net_write: struct {
+        op: ev.NetWriteMsg,
```
zacoons: `NetWriteMsg` doesn't exist. I was hoping this would be an easy change so I could test out zio in my DNS proxy, but it looks like implementing `NetWriteMsg` would be complicated. So I figured I'd post what I've done so far in case you find it useful.

zacoons: Actually, I took another look, and it seems like the pattern is to use `NetSend`.

src/io.zig
```diff
@@ -871,9 +933,25 @@ fn extractBatchResult(data: *BatchCompletionData, tag: Io.Operation.Tag) Io.Oper
                 break :blk .{ null, 1 };
             },
         },
+        .net_send => .{
+            .net_send = blk: {
+                const result = data.net_send.op.getResult() catch |err| break :blk .{ sendMsgErrToSendErr(err), 0 };
+                // Populate the message buffer with received data
+                data.net_send.messages.* = .{
+                    .address = zioIpToStdIo(zio_net.Address.fromPosix(&data.net_send.addr_storage.any, data.net_send.addr_len).ip),
```
zacoons: std.Io requires this to be an `*const IpAddress`, so I'm not sure what to do here.

src/io.zig
```diff
@@ -2468,7 +2546,28 @@ fn sendErrToWriteErr(err: ev.NetSend.Error) Io.net.Stream.Writer.Error {
...
+fn sendMsgErrToSendErr(err: ev.NetSendMsg.Error) Io.net.Socket.SendError {
+    return switch (err) {
+        error.Canceled => error.Canceled,
+        error.SystemResources => error.SystemResources,
+        error.ConnectionResetByPeer => error.ConnectionResetByPeer,
+        error.AccessDenied => error.AccessDenied,
+        error.NetworkDown => error.NetworkDown,
+        error.NetworkUnreachable => error.NetworkUnreachable,
+        error.MessageTooBig => error.MessageOversize,
+        error.WouldBlock,
```
zacoons: It seems like `WouldBlock` shouldn't be an error by this point

lalinsky:
Yes, this is indeed hard, it requires allocations for the temporary conversions. `std.Io.Uring` works around these problems for file writes by degrading to always writing just the first buffer in vector, but that's wrong, that means way more syscalls than necessary. So there will need to be special cases, when we get back with just one write (as is common for the timeout case), we can keep things on stack and no allocation is needed. If this is a true batch, we will need to allocate things.

lalinsky:
I also don't want to spend too much time on designing this, because Jacob Young (who wrote the first working `std.Io.Uring`, including the hack I mentioned) said on zullip some time ago, that he would like to change the ops to have the splat handling on the reader/writer level and make the I/O ops simpler, so this will likely still change if the guys get to finishing the evented impl.

lalinsky:
merged manually as https://github.com/lalinsky/zio/commit/69b266fa2e18810b982d02a6709443dd67a0e906

the zig-0.17 is being rebased, not merged,

zacoons:
cool, thanks!

Issue #668, opened 2026-08-15
Expose a way to wait on an ev.Completion, so a socket read can be raced against a wakeup


nevindra:
There is no supported way to wait on an `ev.Completion` from a fiber, which means a fiber sitting in a socket read cannot also be waiting on anything else. Everything needed to build it is already public except the last step.

## What I am trying to do

Serve a WebSocket connection where the fiber that owns the socket also receives messages sent to it by *other* connections — a chat broadcast. The natural shape is one wait that ends on whichever comes first:

- the socket became readable, or
- something arrived in this connection's mailbox.

The fiber that owns the socket then does all the writing, so nothing writes into a connection it does not own.

## What is already there

`zio.ev` is exported and has each piece:

- `ev.NetPoll(.recv)` — a completion for "this socket is readable"
- `ev.Async` with a thread-safe `notify()` — a completion another fiber can fire
- `ev.Group.init(.race)` — first one to finish wins, the rest are cancelled

That is the same combination `common.timedWaitForIo` uses internally to race a read against a timer, so the pattern is clearly intended to work.

## What is missing

`common.waitForIo(c: *ev.Completion)` is `pub`, but `common.zig` is not re-exported: `zio.zig` takes `Cancelable`, `Timeoutable` and `blockInPlace` from it and nothing else. So a caller can construct the completions and build the race group, and then has no way to park the fiber on the result.

As far as I can tell this is one line in `zio.zig` — though if the intended surface is something narrower than raw `waitForIo` (a `Future` over an `ev.Group`, say, so `select` composes with it), that would suit this use case just as well or better.

## Why it is worth asking for

Without it, the only working shape is a second fiber per connection, one that owns the writes and blocks on the mailbox. I measured that on the framework I am building this for: **8,673 bytes per idle connection**, at 400 connections, RSS delta, repeated until byte-identical. That is roughly the entire per-connection budget a second time, so broadcast is currently not shipped rather than shipped at twice the memory.

`BroadcastChannel` does not avoid it — a shared ring still needs a consumer, and a consumer is a fiber. (It also has a separate problem under cancellation, filed alongside this.)

Happy to test a patch against the real workload if that is useful.

lalinsky:
There is zio.CompletionQueue, which was meant for exactly the use case you are mentioning. I think it might be easier to use that instead it waitForIo. That say, it might be reasonable to export waitForIo, but maybe not directly under the zio namespace.

lalinsky:
[...]

PR #671, opened 2026-08-15
completion_queue: document what CompletionQueue is actually for

lalinsky:
Prompted by #668, where a user wanting to wait on a socket and a cross-fiber wakeup at once concluded the primitive was missing and asked for `waitForIo` to be exported. `CompletionQueue` already does it with public API only — I verified the exact shape works — but nothing in its docs suggested that, because the example was two timers.

## Change

Leads with the choice between the three waits, which is the question a caller actually has:

- `waitForIo` — one operation
- `ev.Group.init(.race)` — several collapsed into one; first wins, **rest cancelled**
- `CompletionQueue` — several kept individually; **the others stay armed**

Then three worked scenarios: one fiber woken by both its socket and its peers (#668), fan-out with per-result handling, and one deadline across a whole batch.

## Tests

The claims the docs lean on are carried by tests so they can't rot quietly:

- `re-submitting the completion that fired leaves the others armed` — the property that separates this from `ev.Group(.race)`
- `one deadline can bound a whole batch` — 501.77ms runtime, i.e. the budget covered the loop rather than restarting per call
- `CompletionQueue: one fiber woken by both its socket and its peers` in `net.zig`, where `Stream` lives. Sequenced rather than raced, so it asserts the re-arming contract instead of a coin flip.

618/618, `zig fmt --check` clean, and the new tests ran 8x with no flakes.

## Two findings while writing this

**Re-arming means re-initialising.** My first drafts of two examples re-submitted a completion without re-initialising it. That does not work: `submit` of a finished completion queues nothing and the next `wait` reports the queue empty rather than blocking. Silent, and it strands the loop. Documented, and both examples corrected.

**Worse, the queue does not survive it.** After such a submit, that completion cannot be armed again even after re-initialising, and `cq.cancel()` then panics with `attempt to use null value` — so a `defer cq.cancel()` turns a caller's mistake into a crash during teardown. I wrote a test pinning this and removed it again rather than land a red suite, since fixing it is a separate change from documenting the primitive. Reproduce with:

```zig
var cq = CompletionQueue.init();
defer cq.cancel();              // panics
var t = ev.Timer.init(.{ .duration = .fromMilliseconds(5) });
cq.submit(&t.c);
_ = try cq.wait();
cq.submit(&t.c);                // finished completion
_ = try cq.wait();              // returns null
```

Worth its own issue; happy to file or fix if you want it.

## Not covered

Doc and test only, no behaviour change. Does not address the export question in #668 — with `CompletionQueue` covering the case, exporting `waitForIo` looks unnecessary, but that is your call to make on the issue.

### DIFF
--- a/src/completion_queue.zig
+++ b/src/completion_queue.zig
@@ -1,28 +1,115 @@
 // SPDX-FileCopyrightText: 2025 Lukáš Lalinský
 // SPDX-License-Identifier: MIT
 
-//! A queue for waiting on multiple I/O operations with an iterator-like interface.
-//!
-//! Unlike `waitForIo` (single operation) or `ev.Group` (combine into one virtual completion),
-//! `CompletionQueue` lets you submit multiple operations, dynamically add more, and process
-//! completions one at a time as they finish.
+//! Wait on several I/O operations at once and handle them one at a time, in
+//! the order they finish.
 //!
 //! Runtime-only: must be used from within an async task context.
 //!
-//! Usage:
-//! ```zig
-//! var cq = CompletionQueue.init();
+//! ## Which of the three waits you want
 //!
-//! var timer1 = ev.Timer.init(.{ .duration = .fromMilliseconds(100) });
-//! var timer2 = ev.Timer.init(.{ .duration = .fromMilliseconds(200) });
-//! cq.submit(&timer1.c);
-//! cq.submit(&timer2.c);
+//! - `waitForIo(&c)` - one operation, block until it finishes.
+//! - `ev.Group.init(.race)` - several operations collapsed into a single
+//!   virtual completion. The first to finish wins and **the rest are
+//!   cancelled**. Use it when the losers are worthless once you have a winner.
+//! - `CompletionQueue` - several operations kept individually. `wait` hands
+//!   back whichever finished, and **the others stay armed**. Use it when each
+//!   operation matters on its own, or when you want to keep waiting after the
+//!   first one lands.
+//!
+//! That last difference is the one that decides most designs. A completion is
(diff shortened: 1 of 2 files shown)


lalinsky:
Closing for now; may be redone later. The branch `completion-queue-docs` stays on origin.

Recording what a redo should change, so it does not start from the same draft:

**The re-arming paragraph documents #673 rather than the contract.** These sentences are that bug's symptom list, not a promise the type makes:

> Handing `submit` a completion that has already finished queues nothing, and the next `wait` reports the queue as empty rather than blocking. There is no recovering from it either: re-initialising that completion afterwards does not bring it back.

Once #673 is fixed the last clause is simply false. The rule is timeless, the failure mode is not, so the doc should say only: a completion carries its result, re-initialise before every submit, and submitting a finished one is a programming error. What goes wrong today belongs on #673.

**Neither new test carries the claim it was written for.**

- `re-submitting the completion that fired leaves the others armed` ends on `cq.hasPending()`, but the loop re-submits `fast` on its last iteration, so `pending` is non-empty because of `fast` regardless of whether `slow` survived. It should wait once more and assert it gets `&slow.c`.
- `one deadline can bound a whole batch` uses three 5ms timers against a 500ms deadline and asserts `seen == 3`. A duration renewed per call gives 3 as well, so it cannot distinguish the two. Staggering to ~100/200/600ms separates them: absolute gives 2, renewed gives 3.

**Minor:** a comment refers to a `cq.cancel()` "below" that is not in its snippet, and the fan-out example omits the `defer cq.cancel()` that the first example and the Teardown section establish.

The Teardown section itself was checked against the code and is accurate. #668 is still worth answering separately; CompletionQueue does cover that case.

Issue #643, opened 2026-08-02
Darwin kqueue wall timers are heavily coalesced and realtime deadlines do not track clock changes


lalinsky:
## Summary

The Darwin kqueue backend arms boot and realtime wall timers with native `EVFILT_TIMER` events. Nominal 100 ms timers regularly arrive 50–150 ms late on GitHub macOS runners, occasionally exceeding the current 250 ms test bound. The realtime path also assumes semantics that XNU does not provide: `NOTE_ABSOLUTE` calendar deadlines are converted to a Mach deadline once at registration and are not subsequently rebased when wall time changes.

This is independent of the Linux backend work in #642.

## CI evidence

Across 15 recent successful `macos-latest` ARM64 release jobs, comparing both backends within each same runner job:

| backend | boot 100 ms | realtime 100 ms |
|---|---:|---:|
| kqueue | mean 162.32 ms, range 100.46–248.34 ms | mean 166.77 ms, range 113.59–235.95 ms |
| poll | mean 102.23 ms, range 100.23–106.74 ms | mean 103.17 ms, range 101.26–107.38 ms |

Observed kqueue failures include:

- boot: 250.20 ms, 250.27 ms, and 465.89 ms
- realtime: 417.99 ms

A rerun of the #642 failure passed, but kqueue still took 171.29 ms for boot and 241.24 ms for realtime; poll took 102.15 ms and 102.20 ms in the same job.

## Root cause

`src/ev/backends/kqueue.zig::armWall` uses:

- boot: `NOTE_NSECONDS | NOTE_MACH_CONTINUOUS_TIME`
- realtime: `NOTE_NSECONDS | NOTE_ABSOLUTE`

Neither uses `NOTE_CRITICAL`. XNU implements `EVFILT_TIMER` via a `thread_call` with `THREAD_CALL_DELAY_USER_NORMAL` unless `NOTE_CRITICAL` is present. Its timer slack is derived from thread QoS and current timer-coalescing policy. This accounts for the large variable latency, while the ordinary kqueue poll timeout remains close to 100 ms.

Primary references:

- XNU timer flags: https://github.com/apple-oss-distributions/xnu/blob/f6217f891ac0bb64f3d375211650a4c1ff8ca1ea/bsd/sys/event.h#L304-L326
[...]

Issue #525, opened 2026-06-26
epoll/kqueue: deferred-send path does ~1 epoll_wait per chunk, capping single-stream throughput (fairness vs batching)


lalinsky:
## Summary

With the single-owner socket registration backend (PR #511 / the persistent edge-triggered path), optimistic **reads complete inline** but optimistic **sends are deferred** (via the deferred-completion queue). Deferring the send turns every successful `writeAll` into a yield point — the fiber parks, the loop ticks, and the tick does a (zero-timeout) `epoll_wait` before processing the deferred completion. So a send-bound fiber pays **~1 `epoll_wait` per chunk**.

This is the right default for request/response and for concurrent servers, but it caps throughput for a **single bulk stream on one executor**. It's a fairness-vs-batching trade, not a correctness bug — filing to track the options.

## Evidence

One-directional stream, 1 executor, 1 connection, 16 KB chunks, 128 MB (`tcp-pingpong-bench ... stream`):

| variant | GiB/s |
|---|--:|
| defer-send (current) | 1.2 |
| inline-send (experiment) | 3.0 |

`strace -c`, 8192 chunks:

| | defer-send | inline-send |
|---|--:|--:|
| `sendto` | 8192 | 8254 |
| **`epoll_pwait2`** | **8199 (~1/chunk)** | **69 (total)** |

`inline-send` doesn't return to the loop per send: it runs `send → send → …` until one hits `WouldBlock` (socket buffer full), so a whole batch costs one `epoll_wait`. `defer-send` ticks the loop (→ `epoll_wait`) after every chunk.

**The gap is bounded to the low-concurrency single-stream case.** With 8+ connections or multiple executors it disappears, and `inline-send` actually *regresses* there (it monopolizes the executor during a send burst, starving other connections — the loop never ticks to service them). 64 KB-chunk streaming, current vs io_uring: ~2.9 vs 3.3 GiB/s at 1 conn, ~8.0 vs 7.7 GiB/s at 4 exec / 32 conn (already on par). So the default is correct under load; only the lone bulk stream pays.

## Root cause

- **defer-send**: a yield point per send → fair (loop services every other fd promptly) but a `epoll_wait` syscall per chunk.
- **inline-send**: batches until backpressure → fast for one stream but monopolizes the loop (unfair), which is why it loses under concurrency and breaks the cooperative-fairness model.

[...]

Issue #495, opened 2026-06-19
Reading from stdin blocks every task on windows


antlanc7:
This code doesn't work on Windows:  stdin_task blocks the other task.

it works perfectly on linux (WSL).

It works if I use `spawnBlocking` but then I get `panic: "Async operations not supported in blocking mode (requires event loop)"` if I do something not trivial (not shown in this example).

```zig
const std = @import("std");
const zio = @import("zio");

fn aux_task() !void {
    while (true) {
        std.log.info("aux_task", .{});
        try zio.sleep(.fromSeconds(1));
    }
}

fn stdin_task() !void {
    var stdin_buffer: [4096]u8 = undefined;
    var stdin_reader = zio.stdin().reader(&stdin_buffer);
    while (true) {
        const line_raw = try stdin_reader.interface.takeDelimiterInclusive('\n');
        const line = std.mem.trim(u8, line_raw, &std.ascii.whitespace);
        std.log.info("got stdin line: {s}", .{line});
        if (std.mem.eql(u8, line, "exit")) {
            break;
        }
    }
}

pub fn main(init: std.process.Init) !void {
    const rt = try zio.Runtime.init(init.gpa, .{
        // .executors = .auto,
    });
    defer rt.deinit();

    var aux_task_handle = try rt.spawn(aux_task, .{});
    defer aux_task_handle.cancel();

    var stdin_task_handle = try rt.spawnBlocking(stdin_task, .{});
    try stdin_task_handle.join();
}
```

Issue #391, opened 2026-05-15
epoll: udp send block the event loop


gBillal:
Consider the following code:
```zig
const std = @import("std");
const Io = std.Io;

const zio = @import("zio");

const dest = Io.net.IpAddress{ .ip4 = .unspecified(9000) };

pub fn main(init: std.process.Init) !void {
    var r = try zio.Runtime.init(init.gpa, .{
        .executors = .exact(2),
    });
    defer r.deinit();

    const io = r.io();

    var grp: Io.Group = .init;
    defer grp.cancel(io);

    var sockets: [10]Io.net.Socket = undefined;
    const port: u16 = 8000;

    for (0..10) |idx| {
        const addr = Io.net.IpAddress{ .ip4 = .unspecified(@intCast(port + idx)) };
        sockets[idx] = try addr.bind(io, .{ .mode = .dgram, .protocol = .udp });
        try grp.concurrent(io, sendData, .{ io, &sockets[idx] });
        try grp.concurrent(io, recvData, .{ io, &sockets[idx] });
    }

    try grp.await(io);
}

fn sendData(io: Io, socket: *const Io.net.Socket) !void {
    while (true) {
        std.log.info("Sending from: {f}", .{socket.address});
        socket.send(io, &dest, "hello world") catch return;
        try io.sleep(.fromMilliseconds(100), .awake);
    }
}

fn recvData(io: Io, socket: *const Io.net.Socket) !void {
    var buffer: [1024]u8 = undefined;

    while (true) {
        std.log.info("Receiving from: {f}", .{socket.address});
        _ = socket.receive(io, &buffer) catch return;
    }
}

```

Running this code will block the event loop and messages are not sent at all. It seems that the socket is bound in blocking mode. Looking at the code, I changed the `netBindIpImpl`:
```zig
fn netBindIpImpl(_: ?*anyopaque, address: *const Io.net.IpAddress, options: Io.net.IpAddress.BindOptions) Io.net.IpAddress.BindError!Io.net.Socket {
    const zio_addr = stdIoIpToZio(address.*);

    var open_op = ev.NetOpen.init(.fromPosix(zio_addr.any.family), .fromStd(options.mode), if (options.protocol) |p| .fromStd(p) else .ip, .{
        .nonblocking = true, // set nonblocking to true
    });

   ....
```

And this resolved the issue and it works as expected.

PR #686, opened 2026-08-18
runtime: poll timeouts on a dedicated executor

### LINKED ISSUE #685: Timers starve when every worker is inside a task that never waits
## What happens

`zio.sleep` (and `select` timeouts / `AutoCancel`) arm on the sleeper's home executor loop. `tick_target_ns` is supposed to force a poll so a self-looping task cannot starve timers, but that only runs if `coro.step()` returns. A task that never waits never returns, so that loop never `poll`s and the timer never fires.

The idle executor does not save this. Timers are registered on the sleeping task's home kqueue (`Waiter.timedWaitClock` → `task.getExecutor().loopSetTimer`). The idle executor sits on a different kqueue and has nothing to fire.

This is not #643 (Darwin kqueue wall-timer coalescing). It reproduces on a monotonic 10ms sleep with CPU hogs that never enter the run loop.

`maybeYield()` still works, because it re-enters `Executor.run`. A tight `Queue.get(min=0)` miss does not: it is a hog that looks like it is doing I/O.

## Repro

Two executors, `enable_main_executor = false`, `enable_task_migration = false` (pinned tasks, the harder case). One 10ms sleeper, then two tasks that spin without waiting. In 200ms the sleeper should tick ~20 times; without a fix it only counts the settle window (~3).

```zig
const runtime = try zio.Runtime.init(allocator, .{
    .executors = .exact(2),
    .enable_main_executor = false,
    .enable_task_migration = false,
});
// spawn sleeper: loop zio.sleep(10ms); ticks += 1
// wall-sleep 30ms so the first timer is armed
// spawn two hogs: while (!stop) { n +%= 1; doNotOptimizeAway(&n); }
// wall-sleep 200ms, stop, join
// ticks >= 10  // fails today; starved at ~3
```

One hog (`spin1`) is a race: starve iff the hog lands on the executor that owns the timer. Two hogs always starve after settle.

## Why I care

my project's mixed SSE + oneshot bench (32 tickers + oneshots on one TLS connection, `exact(2)`, migration off) sits next to Go on oneshot p50 after we stopped parking the connection actor on every complete response. SSE cadence then collapsed: Go at `GOMAXPROCS=2` still hits 16,000 events / 5s; we were missing about half. The actor's empty `tryGet` is the empty-queue hog; the SSE handlers are the sleepers.

## Proposed shape (see following PR)

A dedicated timer executor thread, not counted in `exact(N)` and not a steal victim. Arm timeouts there (`Loop.setTimer` + `wake`, which are already the thread-safe pair). When the timer fires, run the waiter on that thread so it is not queued behind the hog — including with migration off, which would otherwise `scheduleTaskRemote` onto the wedged home.

A sleeper that then becomes a CPU hog on the timer thread can still starve later timeouts; that is the same class as running unbounded work after a sleep, and needs preemption rather than this change.

dimenus:
Fixes #685.

## Why

`tick_target_ns` only forces a poll once `coro.step()` returns. A task that never waits freezes every timer armed on that loop. The idle executor sits on a different kqueue and has nothing to fire.

## What

- Multi-threaded runtimes start one extra **timer executor** thread. It is not in `executors.items`, so `exact(N)` still means N workers and steal/spawn never land on it. It never publishes an idle bit and never steals (stealing a hog onto it would re-starve timers).
- `sleep` / `select` timeouts / `AutoCancel` arm with `Loop.setTimer` + `wake` on that thread (the already-thread-safe pair; `Loop.add` is not).
- A timeout callback sets `timer_wake` so `scheduleTask` runs the waiter **on the timer thread**, including with `enable_task_migration = false`. An ordinary signal from a sleeper already running there does not pull unrelated waiters onto it.
- Builds compiled without task-migration support keep the old home-loop wake (cannot re-home `parent_context_ptr`).

## Test

`runtime: sleep keeps ticking when every worker is spinning` — two hogs, migration off, 10ms sleeper, 200ms window. Starves at ~3 ticks on main; requires ≥ 10.

`./check.sh --full`: 632 passed, 9 skipped.

## Residual

A sleeper that then becomes a CPU hog on the timer thread can still starve later timeouts. That needs preemption, not another courtesy yield in the caller.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -7,8 +7,8 @@ All notable changes to this project will be documented in this file.
 - Sleep, select timeouts, and `AutoCancel` now arm on a dedicated timer executor
   thread, not on the sleeper's home loop. `tick_target_ns` only forces a poll
   once `coro.step()` returns, so a task that never waits used to freeze every
-  timer on that executor; the idle executor sat on a different kqueue and had
-  nothing to fire. The extra thread is not counted in `exact(N)` and is not a
+  timer on that executor; the idle executor sat on a different event loop and
+  had nothing to fire. The extra thread is not counted in `exact(N)` and is not a
   steal victim. A timeout wake runs the waiter there so it is not queued behind
   a hog (including with `enable_task_migration = false`). Builds compiled
   without task-migration support keep the old home-loop path for the wake.
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -2,6 +2,17 @@
 
 All notable changes to this project will be documented in this file.
 
+## [Unreleased]
+
+- Sleep, select timeouts, and `AutoCancel` now arm on a dedicated timer executor
+  thread, not on the sleeper's home loop. `tick_target_ns` only forces a poll
+  once `coro.step()` returns, so a task that never waits used to freeze every
+  timer on that executor; the idle executor sat on a different kqueue and had
+  nothing to fire. The extra thread is not counted in `exact(N)` and is not a
+  steal victim. A timeout wake runs the waiter there so it is not queued behind
+  a hog (including with `enable_task_migration = false`). Builds compiled
+  without task-migration support keep the old home-loop path for the wake.
+
 ## [0.17.0] - 2026-08-05
 
 - Work-stealing is fully wired up now. Idle executors publish an `idle_mask` and coordinate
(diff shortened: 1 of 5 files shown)


dimenus:
Related-issue map is on #685 (not #643 / #199 / #391 / #495; leftover of the “thread doing ticks” note on #525). Changelog no longer says kqueue for a backend-neutral starve.

lalinsky:
I'm sorry, but I'll not accept this. Zio is a cooperative system, you need to yield from your tasks or use spawnBlocking. There are many more things that won't survive blocking task on the executor. I plan to add event loop level stealing, but i consider this a hack that only solves one particular issue.

PR #702, opened 2026-08-21
select: keep the conservation law (cancel-path delivery, claim-before-consume protocol)

### LINKED ISSUE #700: select: a canceled select drops an item a sender already committed to it
## What happens

A task canceled while parked in `select()` returns `error.Canceled` even when a queuing future has already committed a result to it. For a channel branch, the sender's `tryClaim` on the select waiter succeeds any time no winner is set, including while the cancel is landing, and the claim copies the item into the select's stack context. The cancel-path cleanup absorbs the in-flight signal and then discards the result. The item is gone: it is neither in the channel nor returned.

`Channel.receive` already keeps that promise for the same race: on cancellation it absorbs the in-flight signal and returns the item instead of `error.Canceled`. `select` breaks its own protocol comment ("false: ... The future has committed to waking this waiter").

In my project the items carry ownership (write acks that release byte accounting), so every lost item was a leak at connection teardown; the suite hit the accounting assert in about 2 of 9 runs until I traced it here.

## Repro

Race a `trySend` against `cancel()` of a task parked in `select` on that channel plus a never-set `ResetEvent`. Conservation law: after the waiter task ends, the item is EITHER still in the channel OR was returned by the select. Against current `main` it fails at iteration 0.

```zig
// 2000 iterations of:
var ch = Channel(u64).init(&buf);
var never = ResetEvent.init;
var w = try rt.spawn(waiterTask, .{&sh});   // select(.{ .item = ch.asyncReceive(), .never = &never })
while (!sh.parked) try yield(); try yield();
var snd = try rt.spawn(senderTask, .{&sh}); // ch.trySend(1)
w.cancel(); w.join() catch {}; snd.join() catch {};
const in_channel = ch.tryReceive() catch 0;
if (in_channel == 0 and sh.received == 0) return error.ItemDropped; // fires at iteration 0 today
```

The full test is in the PR that follows (`select: canceled select delivers a claimed channel item`).

## Proposed shape (see following PR)

The cancel path runs its own epilogue instead of falling through the shared defer: cancel every registered branch (a claim can still land during this loop, so the winner decision must come after it), absorb every in-flight signal, then return a claimed winner's result if one exists, otherwise `error.Canceled`. The task's cancellation stays pending and re-fires at the next cancelable operation, the same semantics `Channel.receive` has.

Known same-class gap not covered by that change: `selectAwaitables` (the `std.Io.selectImpl` path) has the same cleanup shape, but it returns only an index and result extraction belongs to the caller, so the fix belongs in that caller's layer.

### LINKED ISSUE #701: select: the registration fast path overwrites an earlier branch's claim (item lost, signal miscount)
## What happens

`select()`'s registration loop trusts a boolean `asyncWait`: false means "already complete", and select then does a blind `winner.store(i)` (src/select.zig, the `if (!waiting)` arm). Branches registered earlier in the same pass are already claimable, so a concurrent producer can claim one (winner CAS, item copied into the select's stack context, signal sent) and the store overwrites that claim. Three things go wrong at once:

1. select returns the later branch, and the item already copied for the earlier branch dies on the dead frame;
2. cleanup's signal accounting is off: channel cancel returns `!waiter.didWin()` once the node is out of the queue, and after the overwrite the earlier waiter no longer "won", so cleanup subtracts a signal that is still in flight;
3. that signal can then land on a dead stack waiter.

In my project this dropped inbound wire chunks and write acks at a measured 8-in-30 full-suite runs (a diagnostic CAS panic in the fast path made it loud); the visible tail was a small outbound-accounting leak at connection teardown.

## Repro

Branch `.a` (channel A, empty) registers first; branch `.b` (channel B, already buffered) takes the fast path. A sender on another executor sends into A during the registration pass. Conservation law: the item is EITHER still in A OR returned by the select. On current `main` it fails within a few dozen rounds once idle filler branches between `.a` and `.b` stretch the registration window and the sender sweeps its arrival phase (`round % 64` spins); the full test is in the PR that follows (`select: fast path does not clobber an earlier branch's claim`). Without the fillers the window is tens of nanoseconds and a naive loop misses it for whole runs, which is why it hid for so long.

## Proposed shape (see following PR)

A boolean cannot express "lost, nothing registered", and a prefix `registered_count` breaks once a lost branch can sit between two queued ones. So:

- `asyncWait` returns `.ready | .queued | .lost` (`common.AsyncWaitState`). `.ready` means the future WON the select (`waiter.tryClaim()`) BEFORE any consuming side effect; `.lost` means another branch already won and the future consumed and registered nothing. Direct waiters always claim, so non-select paths treat `.lost` as unreachable.
- `select()` / `selectAwaitables()` track per-branch state (`untouched | queued | lost`) instead of a prefix count; cleanup cancels exactly the queued set and counts expected signals per branch; a pass that observes `.lost` absorbs the winner's in-flight signal and delivers its result.
- Every implementation moves to claim-before-consume: channel receive/send (buffered take, peer handoff, closed), broadcast channel, CompletionQueue ready path, ResetEvent, Future, Awaitable, JoinHandle, Group, Signal, Notify, Timeout.

This is a protocol break for out-of-tree `asyncWait` implementations, which is why I am asking here rather than only sending the patch.

One boundary I kept loud rather than silent: an unbuffered select-vs-select rendezvous (a select-send racing a select-receive with no buffer) can win a select and then find every peer lost. That composition needs a reversible reservation/commit step for the peer handoff; in the PR it panics with a named message instead of stranding the select. Every consumer I have uses direct senders, whose claims cannot fail. If you want that case supported, I would do it as a second step on top.

dimenus:
Fixes #700. Fixes #701.

## Why

Two conservation breaks in `select()`, both found the same way: an item that a sender had committed to a select vanished, neither in the channel nor returned.

1. A task canceled while parked in `select` returned `error.Canceled` after a sender's claim had already copied the item into the select frame (#700).
2. The registration loop's blind `winner.store(i)` on a fast-path branch overwrote a concurrent claim of an earlier-registered branch; the claimed item died, and cleanup's expected-signal count was wrong because `didWin()` no longer held for the clobbered waiter (#701).

## What

Three commits, in order:

- **select: deliver a claimed result when the selecting task is canceled.** The cancel path runs its own epilogue: cancel every registered branch (a claim can land during this loop, so the winner decision comes after it), absorb every in-flight signal, then return a claimed winner's result; otherwise `error.Canceled`. The cancellation stays pending and re-fires at the next cancelable operation, as `Channel.receive` already does. A `cleanup_done` flag keeps the normal-path defer from double-canceling.
- **Future protocol: win the select before any consuming fast path.** `asyncWait` returns `common.AsyncWaitState` (`.ready | .queued | .lost`) instead of `bool`. `.ready` means the future won the select (`waiter.tryClaim()`) before any consuming side effect; `.lost` means another branch already won and nothing was consumed or registered. `select()` and `selectAwaitables()` track per-branch state instead of a prefix `registered_count`, cancel exactly the queued set, and count expected signals per branch. All implementations move to claim-before-consume: channel receive/send (buffered, peer handoff, closed), broadcast channel, CompletionQueue, ResetEvent, Future, Awaitable, JoinHandle, Group, Signal, Notify, Timeout. The protocol doc in `select.zig` carries the new contract.
- **select: raise comptime branch quota for wide selects.** Needed once a select has more than a handful of branches (the new test has 8).

This is a protocol break for any out-of-tree `asyncWait` implementation: the return type changes. Direct waiters always claim, so non-select `wait()` paths are unchanged in behaviour.

## Boundary kept loud

An unbuffered select-vs-select rendezvous (a select-send racing a select-receive with no buffer) can win a select and then find every peer lost. That composition needs a reversible reservation/commit step for the peer handoff, which this PR does not add; it panics with a named message instead of stranding the select. Every consumer I have uses direct senders, whose claims cannot fail. Happy to do the rendezvous case as a follow-up if you want it supported.

## Tests

- `select: canceled select delivers a claimed channel item` — 2000 iterations, `trySend` raced against `cancel()` of a parked select. On `main` without the first commit: `item dropped at iteration 0`, 3 of 3 runs.
- `select: fast path does not clobber an earlier branch's claim` — two executors, one long-lived sender that sweeps its arrival phase (`round % 64` spins), six idle filler branches between the empty `.a` and the already-buffered `.b` so the registration window is wide enough to hit. On the tree with only the first commit: `item dropped at round 19-27`, 5 of 5 runs. With the protocol commit: 3 of 3 green, ~45ms.
- `./check.sh --full` on the branch tip: all checks passed.

## Residual

`selectAwaitables` (the `std.Io.selectImpl` path) gets the per-branch state tracking from the second commit, but the cancel-path result delivery from the first commit applies only to `select()`: `selectAwaitables` returns an index and result extraction belongs to its caller.

### DIFF
--- a/src/awaitable.zig
+++ b/src/awaitable.zig
@@ -7,6 +7,7 @@ const builtin = @import("builtin");
 const RefCounter = @import("utils/ref_counter.zig").RefCounter;
 const WaitNode = @import("utils/wait_queue.zig").WaitNode;
 const Waiter = @import("common.zig").Waiter;
+const AsyncWaitState = @import("common.zig").AsyncWaitState;
 const GroupNode = @import("group.zig").GroupNode;
 const WaitQueue = @import("utils/wait_queue.zig").WaitQueue;
 
@@ -50,15 +51,16 @@ pub const Awaitable = struct {
     }
 
     /// Registers a waiter to be notified when the awaitable completes.
-    /// This is part of the Future protocol for select().
-    /// Returns false if the awaitable is already complete (no wait needed), true if added to queue.
-    pub fn asyncWait(self: *Awaitable, waiter: *Waiter) bool {
-        // Fast path: check if already complete
+    /// This is part of the Future protocol for select(); see
+    /// `common.AsyncWaitState` for the claim-before-ready contract.
+    pub fn asyncWait(self: *Awaitable, waiter: *Waiter) AsyncWaitState {
+        // Fast path: check if already complete (level state, nothing consumed)
         if (self.waiting_list.isFlagSet()) {
-            return false;
+            return if (waiter.tryClaim()) .ready else .lost;
         }
         // Try to push to queue - only succeeds if awaitable is not complete (flag not set)
-        return self.waiting_list.pushUnlessFlag(&waiter.node);
+        if (self.waiting_list.pushUnlessFlag(&waiter.node)) return .queued;
+        return if (waiter.tryClaim()) .ready else .lost;
     }
 
     /// Cancels a pending wait operation by removing the waiter.
--- a/src/common.zig
+++ b/src/common.zig
@@ -38,6 +38,25 @@ pub const Closeable = error{
 };
 
 /// Sentinel value indicating no winner has been selected yet in select operations
+/// Outcome of a future's `asyncWait` under the select protocol.
+///
+/// The conservation rule this encodes: a future must WIN the select (claim
+/// the shared winner slot via `waiter.tryClaim`) BEFORE any consuming fast
+/// path. A fast path that consumed first and reported ready later let a
+/// blind winner store clobber a concurrent claim of an earlier-registered
+/// branch, and the claimed item died on the select's dead stack frame.
+///
+/// - `.ready`: the operation completed on the fast path AND this waiter won
+///   the select. `getResult` is valid now. Nothing was registered.
+/// - `.queued`: the waiter is registered and will be woken exactly once via
+///   `waiter.wake()`. `asyncCancelWait` must be called if the caller stops
+///   waiting.
+/// - `.lost`: another branch of the same select already won, so this future
+///   consumed NOTHING and registered NOTHING. Only select waiters can lose;
+///   a direct waiter's claim always succeeds, so direct callers treat this
+///   as unreachable.
+pub const AsyncWaitState = enum { ready, queued, lost };
+
 pub const NO_WINNER = std.math.maxInt(usize);
 
 /// Stack-allocated waiter for async operations.
--- a/src/completion_queue.zig
+++ b/src/completion_queue.zig
@@ -314,9 +314,11 @@ pub const CompletionQueue = struct {
         return !self.completed.isEmpty() or (self.closed and self.pending.isEmpty());
     }
 
-    pub fn asyncWait(self: *CompletionQueue, waiter: *Waiter, ctx: *WaitContext) bool {
-        // Fast path: something to report already.
-        if (self.isReady()) return false;
+    pub fn asyncWait(self: *CompletionQueue, waiter: *Waiter, ctx: *WaitContext) common.AsyncWaitState {
+        // The ready path consumes nothing (`getResult`/`next` pop), but the
+        // select must still be won before reporting ready: a branch that
+        // registered earlier may already hold a committed result.
+        if (self.isReady()) return if (waiter.tryClaim()) .ready else .lost;
 
         // Park on the completion futex word.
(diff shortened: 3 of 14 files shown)


lalinsky:
Thank you. I thought I already had all the select races figure out, but yeah, this one I missed. I'm not sure about taking this PR though. The panic is not acceptable. I think I need to do what I was avoiding for a long time and make select proper two-phase commit. That enables three things, first solves the race, clean cancellation, and also gives go-style fairness by selecting the random arm when multiple are ready. I need a bit more time to thinking about. I'll see if it's easier to adapt your PR or make it from scratch, because my change is going to be more invasive, but I think more true to what select does.

PR #721, opened 2026-08-29
Move compile-time config to the root module, add zio_options.scheduling

lalinsky:
## Root-module configuration

Compile-time configuration moves out of build options and into the root module, read the way std reads `std_options`:

```zig
// main.zig
pub const zio_options: zio.Options = .{ .scheduling = .work_stealing };
```

Every zio in a binary (yours, and any library's that depends on zio) now sees one configuration. Passing options as `b.dependency` arguments produced a distinct module per argument set, so a binary could end up with two incompatible `Runtime` types. The `-D` options stay in zio's own `build.zig` as defaults for its test and example builds; a root declaration always wins over them.

`backend` also becomes a real enum instead of a string parsed by comparison chain in `ev/backend.zig`.

## zio_options.scheduling

The `task-migration` build option and `RuntimeOptions.enable_task_migration` collapse into one compile-time setting naming the three modes the runtime actually has:

| | |
|---|---|
| `.single_executor` | one executor, count comptime-known, no executor topology |
| `.pinned` | many executors, a task stays on the one it was spawned on |
| `.work_stealing` | many executors, idle ones steal from busy ones |

They nest, so each step down removes machinery rather than branching around it. The five runtime branches on `enable_task_migration` are gone, and the `pending / n_exec` overflow split now compiles to plain `pending` where the overflow queue is executor-local.

## Breaking: multi-threading is opt-in

`scheduling` defaults to `.single_executor`, matching `RuntimeOptions.executors`, which has always defaulted to a single executor. Without the declaration, `.exact(N)` for N > 1 is a **compile error** naming the declaration to add, and `.auto` resolves to one. `ExecutorCount.exact` takes its count as `anytype` so a literal stays comptime-known and the error lands at the call site.

Compile time is deliberate here. A default that silently started running tasks in parallel would turn correct programs into racy ones, and nothing in Zig would flag it. That is not hypothetical: trying `.auto` as the executors default made `runtime.zig`'s "maybeYield yields once the tick budget is spent" (18 tasks incrementing a plain `usize`) fail 4 runs in 5 under `.pinned`, while passing consistently under `.work_stealing`, where the same race just loses fewer increments.

`Runtime.init` no longer returns `error.TaskMigrationNotCompiledIn`, which had no remaining source.

## Optimizations enabled by the above

- `getNextExecutor` homed every new task through a shared round-robin cursor: an atomic increment plus a division by a runtime executor count, once per spawn. With one executor there is nothing to rotate. Interleaved A/B on a spawn loop, 12 rounds of 20k trivial tasks, medians of 5 pairs: **189ns vs 199ns per spawn, ~5% faster**, every pair favouring the fast path.
- State that exists only to move work between executors is zero-bit where nothing moves, following what `IdleMask` already did: the global overflow queue, the searcher count, the stealable flag, and per-executor the steal PRNG and doze flag. Negligible in bytes against a 183KB `Runtime`, but it also drops a PRNG seeding from every executor's init and two dead stores from the run loop.

## Testing

The discipline is comptime-only, so the non-default modes need their own builds to be covered at all. `check.sh` gains `--scheduling` and CI gains a job for `pinned` and `single_executor` (debug and release). Tests needing more than one executor skip where it is not compiled in.

| build | result |
|---|---|
| default | 678 passed, 4 skipped |
| `-Dscheduling=pinned` | 680 passed, 2 skipped |
| `-Dscheduling=single_executor` | 608 passed, 68 skipped |

Examples build on all three, including the `-fsingle-threaded` canary; `-Dbackend=` epoll/io_uring/poll all build.

Two tests now live only in the pinned build: the private-overflow spill test and `net.zig`'s full-duplex-without-loop-group test.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -110,3 +110,33 @@ jobs:
           done
       env:
         TEST_VERBOSE: 2
+
+  # The scheduling discipline is a compile-time choice, so the pinned and
+  # single-executor paths are only covered by building against them.
+  scheduling:
+    strategy:
+      fail-fast: false
+      matrix:
+        zig_version: ['0.16.0']
+        mode: [debug, release]
+        scheduling: [pinned, single_executor]
+
+    runs-on: ubuntu-24.04
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+
+    - name: Setup Zig
+      uses: mlugg/setup-zig@v2
+      with:
+        version: ${{ matrix.zig_version }}
+
+    - name: Build and test
+      shell: bash
+      run: |
+        ./check.sh --full --scheduling ${{ matrix.scheduling }} \
+          ${{ matrix.mode == 'release' && '--release' || '' }}
+      timeout-minutes: 10
+      env:
+        TEST_VERBOSE: 2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,6 +6,10 @@ on:
     tags:
   pull_request:
 
+# Test-only workflow: nothing here needs to write to the repository.
+permissions:
+  contents: read
+
 jobs:
   test:
     strategy:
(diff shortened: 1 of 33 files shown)


lalinsky:
Superseded by #752.

